PR #11, opened 2015-12-04
Demarcate start and end of assignments with arrows

### LINKED ISSUE #10: Visual indication of beginning and ends of assignments
I noticed today that the calendar appears the same when you have two back-to back assignments as when you have one long one.  Clicking on the assignment in question shows the start and end dates, of course.

Perhaps a little cue like a ↦ on the first date and a ↤ on the last date of the assignment?  Not that important.

dfaulken:
Closes #10.

Still can't decide if it's worth writing view specs for this app.

### DIFF
--- a/app/views/assignments/index.haml
+++ b/app/views/assignments/index.haml
@@ -15,6 +15,11 @@
             - if assignment.present?
               = link_to assignment.user.last_name,
                 edit_assignment_url(assignment)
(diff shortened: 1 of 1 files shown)


werebus:
> Still can't decide if it's worth writing view specs for this app.

I wouldn't worry about it.

PR #13, opened 2015-12-04
Rotation generator interface

### LINKED ISSUE #7: Rotation generator
Would be neat to have this. It would generate several back-to-back week-long assignments based on:

+ Semester start date
+ Semester end date
+ Applicable users
+ Weekday of switch

dfaulken:
Closes #7.

Order is alphabetical by last name (or whatever order they show up in in the view, actually).


werebus:
Two thoughts:

1.  A nice default for the start-date would be "the first day with no assignment yet", rather than "next week" since one might be constructing the rotation a few weeks out.
2.  It might also be nice to be able to specify which person to start with.  Since the number of weeks you generate at a time may not be evenly divisible by the number of people in the rotation, folks at the beginning of the alphabet can end up with more weeks than those at the end.  You could use as a default the first person whose name comes after the last person assigned.

dfaulken:
:zap: @werebus

werebus:
It might still be nice to suggest a user as well.

PR #14, opened 2015-12-07
Correcting validation wording

werebus:
Had some trouble creating a user locally. Turned out that the validation
wording wasn't quite right.


werebus:
Also, if we're ever working on this for "real people", Twillio has a pretty sweet API for this:

https://www.twilio.com/lookup

It takes in an arbitrary phone-ish string and returns the country code and digits.  Also, it returns the number formatted the way someone from that country would expect to see it.

PR #43, opened 2017-04-14
Rosters

Anbranin:
In testing your code I reviewed it as well, so I think it's fine for now. I fixed some things I found. What do you think of the tests?


dfaulken:
Will do, looks like there are some Rubocop issues.

app/controllers/rosters_controller.rb
```diff
@@ -0,0 +1,50 @@
+class RostersController < ApplicationController
+  before_action :find_roster, only: [:destroy, :edit, :update]
+
+  def create
+    roster_params = params.require(:roster).permit(:name)
+    roster = Roster.new roster_params
+    if roster.save
+      flash[:message] = 'roster has been created.'
```
dfaulken: Initial capitalization

app/controllers/rosters_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+      redirect_to rosters_path
+    else
+      flash[:errors] = roster.errors.full_messages
+      redirect_to :back
+    end
+  end
+
+  def destroy
+    @roster.destroy
+    flash[:message] = 'roster and any assignments have been deleted.'
```
dfaulken: Initial capitalization

app/controllers/rosters_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+    @rosters = Roster.all
+  end
+
+  def new
+  end
+
+  def update
+    roster_params = params.require(:roster).permit!
+    if @roster.update roster_params
+      flash[:message] = 'roster has been updated.'
```
dfaulken: Initial capitalization

spec/controllers/assignments_controller_spec.rb
```diff
@@ -59,26 +63,27 @@
...
     end
     it 'finds the correct assignment' do
       submit
       expect(assigns.fetch :assignment).to eql @assignment
     end
     it 'populates a users variable of all users' do
-      user_1 = create :user
-      user_2 = create :user
-      user_3 = create :user
+      user_1 = create :user, rosters: [@roster]
```
dfaulken: Do you think it would be worth writing a spec_helper method like `create_user` which you pass a roster, just to cut down on the repetition between lines here?

spec/controllers/assignments_controller_spec.rb
```diff
@@ -90,25 +95,30 @@
...
            end_date: @end_date,
            user_ids: @user_ids,
            starting_user_id: @starting_user_id
     end
     it 'calls Assignment#generate rotation with the given arguments' do
-      expect(Assignment).to receive(:generate_rotation)
-        .with @user_ids, Date.today, Date.tomorrow, @starting_user_id
+      # remove all other instances of roster so 'any instance' definitely
+      # refers to our @roster instance
+      Roster.where.not(id: @roster.id).delete_all
```
dfaulken: OK but why are there any if we're not creating them here

spec/controllers/rosters_controller_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+    let :submit do
+      post :create, roster: { name: 'Operations' }
+    end
+    context 'without errors' do
+      it 'creates a roster' do
+        expect{ submit }.to change{ Roster.count }.from(1).to(2)
+      end
+      it 'redirects to the index' do
+        submit
+        expect(response).to redirect_to(
```
dfaulken: Why not a single line?

spec/controllers/rosters_controller_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+    end
+    let :submit do
+      delete :destroy, id: @roster.id
+    end
+    it 'deletes the correct roster' do
+      submit
+      expect(Roster.where(id: @roster.id)).to be_empty
+    end
+    it 'destroys the roster' do
+      expect_any_instance_of(Roster)
```
dfaulken: Why not a single line?

spec/controllers/rosters_controller_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+    end
+    it 'renders the correct template' do
+      submit
+      expect(response).to render_template :index
+    end
+  end
+
+  describe 'GET #new' do
+    it 'renders the new template' do
+      expect(get :new).to render_template :new
```
dfaulken: Weird parentheses

spec/controllers/rosters_controller_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+    end
+    context 'without errors' do
+      it 'updates the roster' do
+        submit
+        expect(@roster.reload.name).to eql 'unique'
+        expect(@roster.reload.fallback_user).to eql @user
+      end
+      it 'redirects to the index' do
+        submit
+        expect(response).to redirect_to(
```
dfaulken: Why not a single line?

spec/factories/users.rb
```diff
@@ -5,5 +5,6 @@
     sequence(:spire) { |n| n.to_s.rjust(8, '0') + '@umass.edu' }
     sequence(:email) { |n| "user#{n}@umass.edu" }
     sequence(:phone) { |n| '+1' + n.to_s.rjust(10, '0') }
+    rosters { [FactoryGirl.create(:roster)] }
```
dfaulken: Does one need to say `FactoryGirl`?

spec/factories/assignments.rb
```diff
@@ -1,6 +1,7 @@
 FactoryGirl.define do
   factory :assignment do
-    user
+    roster
+    user { FactoryGirl.create(:user, rosters: [roster]) }
```
dfaulken: Does one need to say `FactoryGirl`?

spec/factories/users.rb
dfaulken: Also, holy brackets, :bat: 👨

spec/controllers/assignments_controller_spec.rb
Anbranin: No, because there can't be any users without rosters.

spec/controllers/assignments_controller_spec.rb
Anbranin: when_current_user_is creates a user. Creating a user creates a roster, since there can never be any users unless there are rosters.

spec/controllers/assignments_controller_spec.rb
dfaulken: I don't understand your answer.

My question was whether it would be worth writing a method e.g.:

```ruby
def roster_user(roster)
  create :user, rosters: [roster]
end
```

So that we could cut down on the repetition here, by calling e.g.:

```
user_1 = roster_user(@roster)
```

Anbranin: Then we'd literally never be calling create :user, except to test that it doesn't work without having a roster. My point is there _are_ no users that aren't roster_users.

dfaulken: And *my* point is that the current code is extremely repetitive.

spec/controllers/rosters_controller_spec.rb
Anbranin: What do you mean by 'weird'

spec/factories/assignments.rb
Anbranin: hmm guess not.

spec/controllers/rosters_controller_spec.rb
dfaulken: Eh, never mind, it's not a big deal and Rubocop didn't flag it. Disregard.

spec/controllers/assignments_controller_spec.rb
Anbranin: And I'm saying there's no good way to fix that.

Anbranin: "These 6 lines are repetitive, therefore let us write more stuff every single time we create a user" doesn't seem like a good argument to me.

Anbranin: That is, there are only a few instances when I want a user to belong to a specific roster, so it's repetitive in that case, but very few.

dfaulken: Okay, well, I felt like I was reading the exact same code a bunch but if that's not the case then whatever.

dfaulken: FWIW

```bash
grep -r "rosters: \[@roster\]" . | wc -l
```

gives me 23.

fermion: IMHO there's something to be said for WET tests cases. There's something to be said for total isolation for each example, even if comes at the expense of repetitive LOC.

That being said, if a macro makes sense that that's a way to go here. I think the danger there is that said macro can (and almost always will) grow moving more and more setup logic out of the individual examples making it harder to understand (in the future) why tests fail.

spec/controllers/assignments_controller_spec.rb
```diff
@@ -81,9 +81,9 @@
       expect(assigns.fetch :assignment).to eql @assignment
     end
     it 'populates a users variable of all users' do
-      user_1 = create :user, rosters: [@roster] 
-      user_2 = create :user, rosters: [@roster]
-      user_3 = create :user, rosters: [@roster]
+      user_1 = roster_user @roster
```
fermion: This is really no better or worse than before :trollface:

spec/controllers/assignments_controller_spec.rb
dfaulken: In principle that's true. This example is pretty trivial isolation. If this 'macro' fails, we've got bigger problems.

.rubocop.yml
```diff
@@ -1,2 +1,5 @@
 require: umts-custom-cops
 inherit_from: .rubocop-no-umts-cops.yml
+
+HasAndBelongsToMany:
```
dfaulken: Personally, I like to add comments explaining why we're disagreeing with Rubocop.

Issue #35, opened 2017-04-03
Access control


dfaulken:
@sherson says in https://github.com/umts/transit-it/issues/677:

We might want to have admin and non-admin users (such that non-admin users can't change something they're not involved with). In theory, non-admin users need to be able to:

+ give their own assignment to another person
+ take an assignment from another person
+ take an assignment on dates that are open
+ edit an assignment's dates to allow for splitting up the assignment

dfaulken:
Do we think we should have one admin or multiple admins per rotation?

dfaulken:
Also, should users who are admins in one rotation not be admins in other rotations?

sherson:
> Do we think we should have one admin or multiple admins per roster?

Having multiple could be handy.

> Also, should users who are admins in one roster not be admins in other roster?

Users who are admins in one roster may or may not be admins in another roster.

sherson:
Hey @Anbranin, @dfaulken pointed out that it probably makes sense to create a swap system for this.

Anbranin:
brainstorming

PR #45, opened 2017-04-17
Move from HABTM to has_many through

dfaulken:
First steps towards doing #35.


dfaulken:
@fermion suggested doing more incremental PRs. Is this too incremental? Or just incremental enough?

PR #46, opened 2017-04-17
Use admin status to determine navbar links

dfaulken:
More progress on #35.

Next steps: locking down doing things to assignments.

As admin in current roster:

screenshot from 2017-04-17 14-05-56

As admin, but in another roster:

screenshot from 2017-04-17 14-06-12

As non-admin:

screenshot from 2017-04-17 14-06-24

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -27,6 +27,6 @@ class User < ActiveRecord::Base
   end
 
   def admin?
-    memberships.any?(&:admin)
+    memberships.any?(&:admin?)
   end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -21,4 +21,8 @@ class User < ActiveRecord::Base
   def proper_name
     "#{last_name}, #{first_name}"
   end
+
+  def admin_in?(roster)
+    memberships.find_by(roster: roster).admin?
+  end
 end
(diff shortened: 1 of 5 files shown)


app/models/user.rb
```diff
@@ -21,4 +21,12 @@ def full_name
   def proper_name
     "#{last_name}, #{first_name}"
   end
+
+  def admin_in?(roster)
+    memberships.find_by(roster: roster).admin?
```
Anbranin: I guess the only thing fishy about having being an admin reliant on there being an existing roster is that any new users that want to use this service will need to have an existing admin create a roster for him and make him an admin of that. Or go through IT.

app/models/user.rb
```diff
@@ -21,4 +21,12 @@ def full_name
...
+  end
+
+  def admin?
```
Anbranin: The only issue I can see with having an 'admin' status relying on the fact that you are an admin of an existing roster would be that any new potential administrators of new rosters would have to go through an existing admin to create their roster for them.

dfaulken: Okay. What alternative workflow might you suggest?

Anbranin: One alternative is having a regular old 'administrator' column on a user itself, and then allow current administrators to edit the users to make them admins.. but.. this seems kind of awful, so carry on.

Issue #44, opened 2017-04-17
No good way to assign users to a new Roster


Anbranin:
Currently, to assign users to a new Roster, one must go to the users of a different Roster and edit them to belong to the new Roster.
An easy fix would be to show _all_ Rosters in the navbar, if you are a person with permission to create rosters and assign users to them (@dfaulken is currently working on access control that could make this a reality). For a regular user, showing only the rosters they belong to would be sufficient.

dfaulken:
> show all Rosters in the navbar

Easy, but perhaps not scalable. Worth noting that the rosters index page will show all rosters in a more usable format (vertically rather than horizontally).

Anbranin:
Well, the "manage users" link only goes to the one currently selected in the Navbar, which is why I suggested that. However, you're right. Perhaps a "manage users" link in the rosters index page that goes to the users of that Roster would be good.

jforgue:
On the New roster page/form there should be either a list or a search that allows you to find existing users and add them to the Roster. Added users could be displayed on a table that has columns for the user's name, a radio button for fallback user, a checkbox for making the user an admin, and a button to remove the user from the roster. The user creating the roster should be added to the table when automatically and their remove button could be disabled unless we want to allow users to make rosters they won't be included in.

PR #47, opened 2017-04-17
Allow adding existing users to rotations

dfaulken:
Addresses #44.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,5 +1,5 @@
 class UsersController < ApplicationController
-  before_action :find_user, only: [:destroy, :edit, :update]
+  before_action :find_user, except: %i(create index new)
 
   def create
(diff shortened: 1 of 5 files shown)


dfaulken:
screenshot from 2017-04-17 15-47-51

app/controllers/users_controller.rb
```diff
@@ -22,9 +22,21 @@ def destroy
 
   def index
     @users = @roster.users
+    @other_users = User.all - @users
     @fallback = @roster.fallback_user
   end
 
+  def transfer
```
Anbranin: 'transfer' makes me think we're moving a user from one roster to another.

Anbranin: maybe just `add_roster`?

Anbranin: eh but that's using the word 'roster' more, so I'm not really hooked on changing this.

PR #49, opened 2017-04-17
Admins can only edit their own rosters

dfaulken:
#35.


dfaulken:
Re: CC issue https://github.com/umts/ft-forms/blob/master/app/controllers/application_controller.rb#L12

PR #57, opened 2017-04-19
Implement access control

### LINKED ISSUE #35: Access control

dfaulken:
Closes #35.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -25,6 +25,25 @@ class ApplicationController < ActionController::Base
     redirect_to :back
   end
 
+  # There are three levels of access:
+  # 1. Regular users
+  # 2. Admins in general (of any roster)
+  # 3. Admins of specifically the current roster
+
+  def require_admin
+    # ... and return is correct here
+    # rubocop:disable Style/AndOr
+    head :unauthorized and return unless @current_user.admin?
+    # rubocop:enable Style/AndOr
+  end
+
+  def require_admin_in_roster
+    # ... and return is correct here
+    # rubocop:disable Style/AndOr
+    head :unauthorized and return unless @current_user.admin_in? @roster
+    # rubocop:enable Style/AndOr
+  end
+
   def set_current_user
     if session.key? :user_id
       @current_user = User.find_by id: session[:user_id]
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -73,7 +73,7 @@ class AssignmentsController < ApplicationController
 
   def update
     ass_params = params.require(:assignment)
-                              .permit :start_date, :end_date, :user_id
+                       .permit :start_date, :end_date, :user_id
     unless @current_user.admin_in?(@roster) || taking_ownership?(ass_params)
       require_taking_ownership and return
     end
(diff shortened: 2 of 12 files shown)


app/controllers/assignments_controller.rb
```diff
@@ -80,4 +89,17 @@ def update
   def find_assignment
     @assignment = Assignment.includes(:user).find(params.require :id)
   end
+
+  def require_taking_ownership
+    flash[:errors] = [<<-TEXT]
+      You may only edit or create assignments such that you become on call.
+      The intended new owner of this assignment must take it themselves.
+      Or, a roster administrator can perform this change for you.
```
dfaulken: Open to suggestions on this text.

app/controllers/rosters_controller.rb
```diff
@@ -1,8 +1,9 @@
 class RostersController < ApplicationController
   # The default scaffold method, not the generic one
   # we wrote in ApplicationController.
-  before_action :find_roster, only: [:destroy, :edit, :update]
-  before_action :validate_admin_in_roster, only: %i(destroy edit update)
+  before_action :find_roster, only: %i(destroy edit update)
+  before_action :require_admin_in_roster, only: %i(destroy edit update)
+  before_action :require_admin, except: %i(destroy edit update)
```
dfaulken: It occurs to me that it would be fine to write this as

```ruby
before_action :require_admin
before_action :require_admin_in_roster, only: %i(destroy edit update)
```

Would you like me to make that change?

app/controllers/users_controller.rb
```diff
@@ -56,4 +58,10 @@ def parse_roster_ids(attrs)
     end.compact
     attrs
   end
+
+  def require_admin_or_self
```
dfaulken: Should this be `require_admin_in_roster_or_self`?

app/controllers/users_controller.rb
```diff
@@ -1,5 +1,7 @@
 class UsersController < ApplicationController
   before_action :find_user, except: %i(create index new)
+  before_action :require_admin_or_self, only: %i(edit update)
+  before_action :require_admin_in_roster, except: %i(edit update)
```
dfaulken: Here too, albeit not quite as clearly, this could be:

```ruby
before_action :require_admin_in_roster_or_self
before_action :require_admin_in_roster, except: %i(edit update)
```

dfaulken:
@Anbranin let me know if you'd like me to flag the test changes which are more involved than just the standard action of defining context blocks and setting current users. There are a few.

app/controllers/application_controller.rb
```diff
@@ -25,6 +25,25 @@ def report_errors(object)
...
+    # ... and return is correct here
+    # rubocop:disable Style/AndOr
+    head :unauthorized and return unless @current_user.admin?
+    # rubocop:enable Style/AndOr
+  end
+
+  def require_admin_in_roster
+    # ... and return is correct here
+    # rubocop:disable Style/AndOr
+    head :unauthorized and return unless @current_user.admin_in? @roster
```
dfaulken: I don't understand what you mean. Can you elaborate on how that's a problem?

Anbranin: It will never be an actual problem, not where it's used. disregard.

PR #67, opened 2017-04-25
Fix Codeclimate issues

dfaulken:
This involves moving to mysql2 and bumping a Ruby version, so some amount of server-admin stuff will be needed to merge this.


app/controllers/users_controller.rb
```diff
@@ -74,11 +79,10 @@ def parse_roster_ids(attrs)
...
-    unless @current_user == @user || @current_user.admin_in?(@roster)
-      # ... and return is correct here
-      # rubocop:disable Style/AndOr
-      head :unauthorized and return
-      # rubocop:enable Style/AndOr
-    end
+    return if @current_user == @user || @current_user.admin_in?(@roster)
+    # ... and return is correct here
+    # rubocop:disable Style/AndOr
+    head :unauthorized and return
```
Anbranin: Can you have two return statements in a row like this? Won't it just return and then never reach the `head :unauthorized` bit?

dfaulken: That's exactly the point, yes.

Anbranin: OH I GET IT sorry

PR #72, opened 2017-04-27
The latest and greatest of Rubocop

dfaulken:


dfaulken:
Codeclimate has an older version of Rubocop than we do.

PR #80, opened 2017-06-15
Update Rails to 5.1

mboneil10:
Not sure if the fallback_locations are appropriate. Also, `rspec ./spec/models/roster_spec.rb:66` is failing, but does not affect the coverage.


mboneil10:
Failure/Error: expect(roster.assignments).to receive(:current).and_return assignment

       (#<ActiveRecord::Associations::CollectionProxy [#<Assignment id: 727, user_id: 2938, start_date: "2017-06-14", end_date: "2017-06-16", created_at: "2017-06-15 16:53:31", updated_at: "2017-06-15 16:53:31", roster_id: 3397>]>).current(*(any args))
           expected: 1 time with any arguments
           received: 0 times with any arguments

mboneil10:
I did the merge conflict thing incorrectly. Trying to fix it. But problems.

app/controllers/application_controller.rb
```diff
@@ -24,7 +24,7 @@ def confirm_change(object, message = nil)
 
   def report_errors(object)
     flash[:errors] = object.errors.full_messages
-    redirect_to :back
+    redirect_back fallback_location: 'public/404.html'
```
dfaulken: Why a 404 page here?

Difficult to come up with a truly generic path here, so I might recommend another method argument.

config/cable.yml
```diff
@@ -0,0 +1,10 @@
+development:
+  adapter: async
+
+test:
+  adapter: async
+
+production:
+  adapter: redis
+  url: redis://localhost:6379/1
+  channel_prefix: screaming_dinosaur_production
```
dfaulken: It may be worth also considering just not adding ActionCable if we don't have a use for it. That is, one could imagine we might have generated a *new* Rails 5.1 screaming-dinosaur with it excluded (`rails new screaming-dinosaur -C ...`).

config/locales/en.yml
```diff
@@ -16,11 +16,18 @@
...
+#   'true': 'foo'
+#
 # To learn more, please read the Rails Internationalization guide
 # available at http://guides.rubyonrails.org/i18n.html.
 
 en:
-  activerecord:
-    attributes:
-      user:
-        is_fallback: Fallback
```
dfaulken: Hey wait, why was this removed?

spec/controllers/assignments_controller_spec.rb
```diff
@@ -18,7 +18,7 @@
       }
     end
     let :submit do
-      post :create, roster_id: @roster, assignment: @attributes
+      post :create, params: { roster_id: @roster, assignment: @attributes }
```
dfaulken: OOC, did you use a macro for this?

spec/controllers/changes_controller_spec.rb
```diff
@@ -22,7 +22,7 @@
         @destroy_version = @destroyed.versions.last
       end
     end
-    let(:submit) { get :undo, id: version.id }
+    let(:submit) { get :undo, params: { id: version.id } }
```
dfaulken: Double ending curly braces

meh favrit

dfaulken: It's fine though, there's not a better way, I just don't like how it looks personally.

app/controllers/application_controller.rb
mboneil10: That's the reason. I thought it was a generic path. 

What do you mean another method argument? Removing the `redirect_back` method?

config/locales/en.yml
mboneil10: No idea.

config/cable.yml
mboneil10: Okay

app/controllers/application_controller.rb
dfaulken: Another argument to `report_errors`, being the fallback location.

config/locales/en.yml
mboneil10: Oh. It happened in the commit where I accepted all overwrites after running `rails app:update`. I was told to accept all overwrites. Is this wrong? I can put it back.

mboneil10: The routes file was also removed... so I had to rewrite it (copied the route file from master).

config/cable.yml
mboneil10: @Anbranin, do you have an opinion on this?

spec/controllers/assignments_controller_spec.rb
mboneil10: What's a macro?

spec/controllers/changes_controller_spec.rb
mboneil10: Okie dokie

spec/controllers/assignments_controller_spec.rb
mboneil10: Ohhh. For vim. Nope. Should have.

config/locales/en.yml
dfaulken: If there's no reason to remove it, yes, I think it should stay. I can also explain what this file is if you're unfamiliar.

mboneil10: Yes please, I am unfamiliar.

dfaulken: In short: sometimes you want your column name to say one thing for the sake of developer clarity, but be called another thing when viewed for the sake of user clarity. The locales file provides a central place to define what those 'translations' should look like. Rails then uses them by default to do things like `label`s, `human_attribute_name`s, etc.

More details at http://guides.rubyonrails.org/i18n.html.

mboneil10: Got it. Thanks!

dfaulken: Removing the `is_` is a trivial example; one from `umts/incidents` is that I have a column `other_vehicle_owned_by_other_driver` which is shorter to type than its translation 'Is the other driver involved the owner of the vehicle?', which is clearer for a person to read than how it would be humanized by default ('Other vehicle owned by other driver?').

app/controllers/application_controller.rb
mboneil10: I'll change the tests accordingly, too.

spec/controllers/assignments_controller_spec.rb
mboneil10: 🐑

app/controllers/application_controller.rb
```diff
@@ -22,9 +22,9 @@ def confirm_change(object, message = nil)
     flash[:message] = message
   end
 
-  def report_errors(object)
+  def report_errors(object, fallback_location)
```
dfaulken: I might suggest a keyword argument, so that it's clear what passing a path means.

config/cable.yml
Anbranin: We do not need ActionCable for this app. I'd exclude it.

app/controllers/assignments_controller.rb
```diff
@@ -44,7 +44,8 @@ def generate_rotation
       flash[:errors] = 'The starting user must be in the rotation.'
       # ... and return is correct here
       # rubocop:disable Style/AndOr
-      redirect_to :back and return
+      redirect_back(fallback_location:
+                    roster_assignments_path(@roster)) and return
```
dfaulken: Shouldn't need quite so many parentheses - I think you can safely lose the outer set.

dfaulken:
Just need to figure out what's going on with that test.

PR #93, opened 2017-08-28
new users fixed

### LINKED ISSUE #92: Creating new user raises exception
user_params:
`<ActionController::Parameters {"first_name"=>"", "last_name"=>"", "spire"=>"", "email"=>"", "phone"=>"", "reminders_enabled"=>"1", "change_notifications_enabled"=>"1", "rosters"=>["", "2"], "membership"=><ActionController::Parameters {"admin"=>"0"} permitted: true>} permitted: true>`

Exception:
Roster(#70242086014780) expected, got "" which is an instance of String(#70242013708040)

avacmeyer:
#92 New users can be made now.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -12,12 +12,13 @@ class UsersController < ApplicationController
 
   def create
     user_params = params.require(:user).permit(*WHITELISTED_ATTRIBUTES)
-    user = User.new user_params
-    user.rosters << @roster
-    if user.save
-      confirm_change(user)
+    membership_params = user_params[:membership]
+    user_params = parse_roster_ids(user_params.except(:membership))
+    @user = User.new(user_params)
+    if @user.save && update_membership(membership_params)
+      confirm_change(@user)
       redirect_to roster_users_path(@roster)
-    else report_errors(user, fallback_location: roster_users_path)
+    else report_errors(@user, fallback_location: roster_users_path)
     end
   end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -36,5 +36,4 @@ class User < ApplicationRecord
   def membership_in(roster)
     memberships.find_by(roster: roster)
   end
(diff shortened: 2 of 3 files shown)


app/controllers/users_controller.rb
```diff
@@ -12,12 +12,13 @@ class UsersController < ApplicationController
 
   def create
     user_params = params.require(:user).permit(*WHITELISTED_ATTRIBUTES)
-    user = User.new user_params
-    user.rosters << @roster
-    if user.save
-      confirm_change(user)
+    membership_params = user_params[:membership]
+    user_params = parse_roster_ids(user_params.except(:membership))
+    @user = User.new(user_params)
```
Anbranin: Why does `@user` need to be an instance variable?

Anbranin:
Also, run rubocop

app/models/user.rb
```diff
@@ -36,4 +36,5 @@ def admin?
   def membership_in(roster)
     memberships.find_by(roster: roster)
   end
+
```
dfaulken: I think Rubocop might object to the presence of this line.

app/controllers/users_controller.rb
avacmeyer: Because we called a method that uses the instance variable.

Anbranin:
Are the test failures not a concern?

app/controllers/users_controller.rb
Anbranin: Should be a local variable.

dfaulken: No, @avacmeyer is correct. [The update membership method](https://github.com/umts/screaming-dinosaur/blob/ef61b663e6a61f224bc3d8888338988b2ce55264/app/controllers/users_controller.rb#L68-L74) uses an instance variable.

Anbranin: oh weird ok

PR #96, opened 2018-01-08
Code Climate: Return of the conditional for variable assignment

### LINKED ISSUE #59: Fix "Rubocop/Metrics/MethodLength" issue in app/models/assignment.rb
Method has too many lines. [12/11]

https://codeclimate.com/github/umts/screaming-dinosaur/app/models/assignment.rb#issue_58f7b7216bb2e4000100004d

jforgue:
Closes #59


app/models/assignment.rb
```diff
@@ -79,11 +79,15 @@ def send_reminders!
...
-                                    .where("start_date <= ? AND end_date >= ?",
-                                           end_date, start_date)
+                                      .where('
+                                      start_date <= ? AND end_date >= ?',
+                                             end_date, start_date)
                               else
-                                roster.assignments.where("start_date <= ? AND end_date >= ? AND id != ?",
-                                           end_date, start_date, id)
+                                roster.assignments
+                                      .where('
```
werebus: These breaks are in weird places.  The reason to do something different depending on `new_record?` is because we want to avoid comparing to `NULL`, which is undefined. Is it more or less bad to do any of these:

```ruby
overlapping_assignments = roster.assignments.where(
  "start_date <= ? AND end_date >= ? AND id != ?",
  end_date, start_date, id.to_i
)
# OR
overlapping_assignments = roster.assignments.where(
  "start_date <= ? AND end_date >= ? AND id != ?",
  end_date, start_date, (persisted? && id)
)
```

Hacky?

jforgue: I see what you mean, I like option #2

jforgue: But what if the assignment is new and the id isn't yet persisted?

werebus: Those are the same thing. `persisted?` is the opposite of `new_record?`

jforgue: I had the wrong impression of what would be returned by the combination of `(persisted? && id)` if id was null. I think I'll go with what you first suggested since that seemed a little more obvious to me

dfaulken:
Oh, no, I'm sorry. I interpreted your PR title to be referring to a Rubocop failure, but it was actually a complexity metric.

Personally, I don't think what's there is too complex. Although I guess I thought differently when I opened the issue. Past me might disagree, but present me votes to leave the code as-is, and to write a comment explaining what's going in if we think it might be unclear.

jforgue:
@Anbranin What do you think?

Anbranin:
I don't actually understand why seeing if something overlaps could have anything to do with whether it's a new record or not. And I don't know what calling `id.to_i` is supposed to do.

werebus:
A non-new record always overlaps itself. So, you want to exclude that from your query.

But you can't do this:

```ruby
roster.assignments.where("start_date <= ? AND end_date >= ? AND id != ?",
                         end_date, start_date, id)
```

with a new record because `id` is `nil`, and you can't compare with `=` or `!=` to `NULL`.  `nil.to_i` is `0`.

That's the "why", but is it a good idea? I'm not sure because everyone who's seen it so far has been confused by it.

Anbranin:
I could be missing something, because in that case for a new record wouldn't overlapping assignments never be blank?

dfaulken:
> for a new record wouldn't overlapping assignments never be blank?

No, it's a new record. It can't be found by a database query yet, because it doesn't exist in the database. That's why we only have to do the ID comparison for persisted records, which would "overlap with themselves" if we don't exclude them.

Anbranin:
Ooooohhhhhhhhhhh I read Matt's comment as "A new record always overlaps with itself" and I was like what
I get it now

dfaulken:
So what do you want to do?

Anbranin:
Leave it as-is and write a comment. Alternatively, write one query and then an if-statement to ask whether the departments include self. Of course, that adds bulk, but then there's no comment needed.
I leave it entirely up to @jforgue.

PR #98, opened 2018-01-09
Remove delete from form and realign

### LINKED ISSUE #97: Saving an edited assignment deletes instead
If you save after editing an assignment, even if there is no change in dates, the assignment will be deleted.

jforgue:
Closes #97 
screen shot 2018-01-09 at 3 03 59 pm


dfaulken:
Does this feature have tests? If so, does this affect them? If not, should we implement some?

jforgue:
This feature does not currently have any tests, but probably should.

jforgue:
These tests are not the final product

spec/features/assignments/edit_spec.rb
```diff
@@ -39,4 +44,67 @@
...
+      select(@last_name, from: :assignment_user_id)
+      click_button 'Save'
+      within('div.alert.alert-danger') do
+        expect(page).to have_selector 'li', text: 'You may only edit'
+      end
+    end
+    it 'allows admin users to edit all assignments' do
+      membership = user.membership_in(roster)
+      membership.admin = true
+      membership.save
```
dfaulken: FYI, could also do `user.membership_in(roster).update admin: true` to replace the above 3 lines

spec/features/assignments/edit_spec.rb
```diff
@@ -39,4 +44,67 @@
...
+      visit roster_assignments_url(roster, date: date_today)
+      visit edit_roster_assignment_url(roster, assignment)
+      select(@last_name, from: :assignment_user_id)
```
dfaulken: Does the selector have a label? If so, preferable to select for it by that, since it's what the user can see, and how a user would be able to tell what the box is for.

spec/features/assignments/edit_spec.rb
```diff
@@ -39,4 +44,67 @@
...
+      click_button 'Save'
+      # Since the assignment is 7 days long
+      expect(page).to have_selector 'a',
+                                    text: @last_name, count: 7
```
dfaulken: Perhaps add something about the `href`, so that you're sure it's not a link to a different assignment belonging to the right person?

spec/features/assignments/edit_spec.rb
```diff
@@ -39,4 +44,67 @@
...
+    end
+  end
+  context 'changing the assignment' do
+    let(:last_name) { user.last_name }
+    it 'updates the assignment' do
+      # Visit the index on the correct month to ensure the form
+      # submit returns us to the correct month
+      visit roster_assignments_url(roster, date: date_today)
+      visit edit_roster_assignment_url(roster, assignment)
+      fill_in('assignment[end_date]', with: date_today)
```
dfaulken: Here too, it'd be better to find the select tag by an identifier that a user could see.

app/views/assignments/edit.haml
```diff
@@ -14,8 +14,9 @@
         .form-group
           = f.label :end_date
           = f.text_field :end_date, class: 'datepicker form-control'
-  .center
-    = f.submit 'Save', class: 'btn btn-primary'
+    .col-xs-offset-4.col-xs-1
+      = f.submit 'Save', class: 'btn btn-primary'
```
dfaulken: Could do this inline, e.g.

```haml
.class= content
```

spec/features/assignments/edit_spec.rb
```diff
@@ -39,4 +44,70 @@
...
+      # submit returns us to the correct month
+      visit roster_assignments_url(roster, date: date_today)
+      visit edit_roster_assignment_url(roster, assignment)
+      fill_in('End date', with: date_today)
+      click_button 'Save'
+      # Since the assignment is now 5 days long
+      expect(page).to have_link last_name,
+                                href: edit_roster_assignment_url(roster.id,
+                                                                 assignment
+                                                                     .id),
```
dfaulken: 🤢 

In this case

```ruby
expect(page).to have_link last_name,
  edit_roster_assignment_url(roster, assignment)
```

would be better.

Note that you don't have to explicitly pass the IDs of the two objects. The objects themselves will do.

PR #100, opened 2018-01-09
Keep the current user selected when editing an assignment.

### LINKED ISSUE #99: Editing an assignment doesn't preserve assigned user
The select box on the form doesn't have a `selected` `option`

werebus:
Fixes #99

### DIFF
--- a/app/views/assignments/edit.haml
+++ b/app/views/assignments/edit.haml
@@ -5,9 +5,8 @@
       .col-xs-2.col-xs-offset-3
         .form-group
           = f.label :user_id
-          = f.select :user_id,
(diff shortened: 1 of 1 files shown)


app/views/assignments/edit.haml
```diff
@@ -5,9 +5,8 @@
       .col-xs-2.col-xs-offset-3
         .form-group
           = f.label :user_id
-          = f.select :user_id,
-           options_from_collection_for_select(@users, :id, :last_name),
-           class: 'form-control'
+          = f.collection_select :user_id, @users, :id, :last_name,
+                                {}, class: 'form_control'
```
dfaulken: Should be `form-control`, I think.

Issue #60, opened 2017-04-21
Destroying an assignment goes to this month


dfaulken:
Even if it was in the future. It'd be nice if it sent you back to where you came from.

Related to #42.

PR #102, opened 2018-01-10
Redirect to index with month you were looking at

### LINKED ISSUE #42: Month-boundary-spanning assignments can return you to the wrong month
Today is Tuesday, April 4th. The currently on call person started their assignment on Friday, March 31st. If you click on their assignment, make a change (or not) and click 'Save', you get sent back to the calendar for March, since that's when that assignment starts.

It'd be nice if we remember which month we were looking at, so that we can return the person to the same month they were in.

jforgue:
Closes #42


coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 100.0
+    "covered_percent": 27.54
```
dfaulken: Whoops

spec/features/assignments/edit_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    set_current_user(user)
+  end
+  after :each do
+    Timecop.return
+  end
+
+  it 'returns the user to the appropriate index page' do
+    date_today = Date.new(2017, 4, 4)
+    start_date = Date.new(2017, 3, 31)
+    assignment.start_date = start_date
```
dfaulken: Why not set the date that way in the `let` block?

spec/features/assignments/edit_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    visit roster_assignments_url(roster, date: date_today)
+    visit edit_roster_assignment_url(roster, assignment)
+    click_button 'Save'
+    expect(current_url)
+      .to eq roster_assignments_url(roster,
+                                    date: date_today
+                                              .beginning_of_month)
```
dfaulken: Probably worth setting that start of month date in a variable beforehand.

spec/features/assignments/edit_spec.rb
jforgue: I set some variables using `let` because I know they'll be used for future tests, I expect to use these specific dates only for this test. I can certainly could put them in a `context` block or under the `describe` block if that would be better?

jforgue: Whoops, don't mind that comment. I was looking at the wrong line.

spec/features/assignments/new_spec.rb
```diff
@@ -13,15 +13,25 @@
...
-    visit roster_assignments_url(roster, date: date_today)
-    visit new_roster_assignment_url(roster, date: start_date)
-    click_button 'Create'
-    expect(current_url)
-      .to eq roster_assignments_url(roster,
-                                    date: date_today
-                                              .beginning_of_month)
+  context 'returns the user to the appropriate index page' do
+    before :each do
+      @date_today = Date.new(2017, 4, 4)
```
dfaulken: `let`s would be preferable to instance variables (here and in general).

spec/features/assignments/new_spec.rb
```diff
@@ -13,15 +13,25 @@
...
+      click_button 'Create'
+      expect(current_url)
+        .to eq roster_assignments_url(roster,
+                                      date: @month_date)
+    end
+    it 'displays the correct month' do
+      visit roster_assignments_url(roster, date: @date_today)
+      visit new_roster_assignment_url(roster, date: @start_date)
+      click_button 'Create'
+      expect(page).to have_selector '.title', text: (@month_date.strftime '%-B %G')
```
dfaulken: Parentheses should be distributed:

```ruby
expect(page).to have_selector '.title', text: @month_date.strftime('%-B %G')
```

That is, further down the syntax tree.

spec/features/assignments/new_spec.rb
jforgue: Right, I knew they were faster but for some reason I thought you couldn't edit an attribute of a `let` unless it was an instance variable. Not sure why though.

jforgue:
Closes #60

PR #105, opened 2018-02-13
Deactivate and reactivate users

### LINKED ISSUE #77: Deactivate/reactive users
Would be useful to be able to deactivate and reactivate users (for example, when a user leaves).

jforgue:
Still need to test, just looking for feedback at this time.
Closes #77


app/controllers/users_controller.rb
```diff
@@ -34,6 +34,13 @@ def index
     @users = @roster.users
     @other_users = User.all - @users
     @fallback = @roster.fallback_user
+    @active = User.active.order :last_name
+  end
+
+  def inactive
```
dfaulken: I would recommend a parameter to the `index` action that controls whether inactive users are shown, rather than a separate controller action and view.

This way, if we would like to make changes to the users index in the future, we don't need to remember to mirror those changes on the 'inactive' pages.

Additionally, I might recommend showing *either* active or inactive users, but not both, though I'm happy to be overruled on that.

app/views/users/inactive.haml
```diff
@@ -0,0 +1,30 @@
...
+      %th Phone
+      %th
+      %th
+  %tbody
+    - @inactive.each do |user|
+      %tr
+        %td
+          = user.full_name
+          - if user == @fallback
+            (fallback)
```
Anbranin: Why do we need the information about fallbacks if they're inactive?

app/controllers/users_controller.rb
```diff
@@ -34,6 +34,13 @@ def index
...
+    @users = @roster.users
```
Anbranin: You're not using this instance variable in that view. You don't need it.

app/models/user.rb
```diff
@@ -16,6 +16,10 @@ class User < ApplicationRecord
   validates :phone,
             format: { with: /\A\+1\d{10}\z/,
                       message: 'must be "+1" followed by 10 digits' }
+  before_save -> { assignments.future.destroy_all }, if: :being_deactivated?
+
+  scope :active, -> { where active: true }
```
Anbranin: You mostly use active when you're querying users - why not just make this a default scope?

app/views/users/index.haml
```diff
@@ -26,7 +28,7 @@
       %th
       %th
   %tbody
-    - @users.each do |user|
+    - @active.each do |user|
```
Anbranin: Not a terrible idea, but if it were me I'd just use `@users` in both views. They're both collections of users--information about what kinds of users they are can remain in the controller.

app/views/users/inactive.haml
dfaulken: Quite. Indeed we should probably enforce that inactive users cannot be the fallback.

app/views/users/index.haml
dfaulken: For an alternative approach, I suggested not having two views.

app/models/user.rb
dfaulken: I actually don't recommend default scopes. As an example, in order to get inactive users, you'd need to write `User.unscoped.inactive`, which I find clumsy and weird.

werebus: Agreed. And if the goal is not confusing future developers, `User.inactive` always returning and empty collection is not ideal. Nor do I particularly like `User.all` returning only active users rather than, you know, "all" of them.

app/controllers/users_controller.rb
```diff
@@ -52,7 +59,7 @@ def update
     if @user.update(user_params) && update_membership(membership_params)
       confirm_change(@user)
       if @current_user.admin_in? @roster
-        redirect_to roster_users_path(@roster)
+        redirect_back fallback_location: roster_users_path(@roster)
```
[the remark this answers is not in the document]
jforgue: Since I was using the update action to handle activation I didn't want to redirect back to the index while making a change on the inactive page. Although, that page is gone now so I can change it back.

spec/features/users/index_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      before :each do
+        @inactive_user = create :user, active: false
+        @membership = create :membership, roster: roster, user: @inactive_user
+        click_link 'Inactive users'
+      end
+      it 'shows active users button' do
+        expect(page).to have_link 'Active users'
+      end
+      it 'shows activate user button on users' do
+        expect(page).to have_selector 'td', text: @inactive_user.first_name
```
mboneil10: What happens when users aren't there?

spec/features/users/index_spec.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'user pages' do
+  let(:roster) { create :roster }
```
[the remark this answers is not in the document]
jforgue: I believe rubocop disagrees, I could be wrong though.

app/controllers/users_controller.rb
```diff
@@ -74,9 +79,11 @@ def update_membership(membership_params)
   end
 
   def parse_roster_ids(attrs)
-    attrs[:rosters] = attrs[:rosters].map do |roster_id|
-      Roster.find_by id: roster_id
-    end.compact
+    if attrs[:rosters].present?
+      attrs[:rosters] = attrs[:rosters].map do |roster_id|
+        Roster.find_by id: roster_id
+      end.compact
```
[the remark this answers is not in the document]
mboneil10: `compact` returns a copy of self with all `nil` elements removed.

spec/features/users/index_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  let(:admin_membership) { create :membership, roster: roster, admin: true }
+  let(:admin) { admin_membership.user }
+  context 'viewing the index' do
+    before :each do
+      set_current_user(admin)
+      visit root_path
```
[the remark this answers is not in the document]
jforgue: The [Capybara documentation](https://github.com/teamcapybara/capybara#gotchas) recommends staying away from absolute urls.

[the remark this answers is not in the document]
werebus: Right. Although it's telling that that configuration option is for Action Mailer. Generally, I think it's better to use `_path` than `_url`. In the same way that I was taught not to type the domain into every `href` when I was learning HTML.

app/controllers/users_controller.rb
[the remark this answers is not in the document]
werebus: Right, although that's after the `.map` block. An array with some `nil`s in it is still present.

However, because you're looking for rosters by id and no `Roster` will have a `nil` `id`, this can all just be done with where:

```ruby
def parse_roster_ids(attrs)
  attrs[:rosters] = Roster.where(id: attrs[:rosters]
  attrs
end
```

spec/features/users/index_spec.rb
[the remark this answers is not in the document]
dfaulken: @mboneil10 you do. When you don't need it is:

```ruby
let :roster do
  create :roster
end
```

But with no parentheses using curly braces you'll get a syntax error - essentially because curly braces are more ambiguous than `do` and `end`.

spec/features/users/index_spec.rb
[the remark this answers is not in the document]
dfaulken: Yeah, you're right. https://github.com/umts/incidents/issues/145

spec/features/users/index_spec.rb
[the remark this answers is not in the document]
mboneil10: Whoops got it!

app/controllers/users_controller.rb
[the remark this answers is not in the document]
mboneil10: +1 for Justin

spec/features/users/index_spec.rb
jforgue: There will be no users displayed in the table. The only visual change is a lack of users and the "Active users" button which is already tested for. I could test for a lack of users if you think that's necessary?

Anbranin:
Hey what happened to the coverage? is this thoroughly tested?

Anbranin:
Oops I'm so used to working on projects that aren't at 100% coverage that I always assume it'll go up when someone writes tests! Disregard.

spec/models/user_spec.rb
```diff
@@ -48,4 +46,13 @@
       end
     end
   end
+  describe 'being deactivated' do
+    let(:future_assignment) { create :assignment, start_date: Date.tomorrow }
+    it 'destroys future assignments for users' do
+      user = future_assignment.user
+      user.active = false
+      user.save
+      expect(user.assignments).to eq []
```
Anbranin: could you use `be_empty` on this?

werebus: Interesting, though, that an empty `ActiveRecord::Relation` actually _does_ `eq []`

PR #106, opened 2018-03-07
Working system tests

Anbranin:
There are some non-functional changes to this PR, such as:
1. Removing `rails_helper.rb` entirely and putting everything in `spec_helper.rb`. There is no functional reason to have these two separate files, since the only time it would make a difference is if we regularly ran, in isolation, tests on modules in the lib directory. I think.
2. Grouping the `:test` gems together so they don't get loaded in dev
3. Removing all the Capybara configuration that actually did nothing, and was merely a previous attempt to get things working to begin with.

The functional change:
1. Bundle updating
2. Adding the gem "puma" to the :test group in the Gemfile.

Let me explain:
The initial error messages caused by the problem this PR solves all boiled down to the same thing: It was as if Capybara's methods just weren't getting loaded into the test. "visit" didn't work, neither did "click_on" or any other of the Capybara::DSL. Many Googlings later and we were no closer to knowing why. Many things were `include`d, and Capybara was configured up the wazoo.
But, running a simple bundle update and then attempting to run the specs in spec/system got me an error message I could use:
```
System test integration requires Rails >= 5.1 and has a hard dependency on a webserver and
 `capybara`, please add capybara to your Gemfile and configure a webserver (e.g. `Capybara.server = 
:webrick`) before attempting to use system tests.
```
Now, this didn't make any sense to me, so I googled this error, this time finding [this page](https://github.com/rspec/rspec-rails/issues/1930).
I added `puma` to the `:test` directory and now system tests work.


Gemfile
```diff
@@ -34,20 +35,23 @@ group :development do
...
   gem 'codeclimate-test-reporter', '~> 1.0'
   gem 'haml_lint'
   gem 'mocha'
   gem 'pry-byebug'
+  gem 'rubocop'
+  gem 'umts-custom-cops'
+  gem 'timecop'
+end
+
+group :test do
```
jforgue: I like this, I thought about doing it myself.

spec/controllers/assignments_controller_spec.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require 'rails_helper'
```
jforgue: I know removing `rails_helper` has no functional difference for us but, the thing that bothers me about this is rails thinks we should have both. Plus from what I've read, mainly [here](https://stackoverflow.com/questions/24145329/how-is-spec-rails-helper-rb-different-from-spec-spec-helper-rb-do-i-need-it), the `rails_helper` should be required for any test that depends on rails and `spec_helper` should be required for those that don't.

spec/system/user_spec.rb
```diff
@@ -1,18 +1,19 @@
 # frozen_string_literal: true
 
-require 'rails_helper'
+require 'spec_helper'
 
-describe User do
```
jforgue: You may want to rename this to match a feature instead of an object like we do with our feature tests. I know I'm the one who named it but I wasn't thinking about what these tests were actually for.

spec/system/user_spec.rb
```diff
@@ -1,18 +1,19 @@
...
-  let(:user) { create :user }
+describe 'user pages' do
   let(:membership) { create :membership, admin: true}
-  let(:admin) { create :user, memberships: [membership] }
+  let!(:admin) { create :user, first_name: 'tester', memberships: [membership] }
 
   context 'viewing the index' do
     it 'directs you to the appropriate page' do
       set_current_user(admin)
       visit root_url
```
jforgue: I might suggest we use `root_path` instead. We talked about using a combination of `Rack::Test` and `Selenium` for better speed. I found in the Capybara [documentation](https://github.com/teamcapybara/capybara#gotchas) that trying to visit absolute URLs can cause issues.

Gemfile
```diff
@@ -16,6 +16,7 @@ gem 'sass-rails'
 gem 'snappconfig'
 gem 'uglifier'
 gem 'whenever', require: false
+gem 'puma', '~> 3.7'
```
jforgue: So is `puma` already the default web server for rails and it's just not included in the gemfile?

Anbranin: It's the default for new rails 5 apps and when this was created, it was in Rails 4 so didn't get it. Obviously `rails s` still works so it's fine.

spec/system/user_spec.rb
Anbranin: Yeah, I want to eventually do that but this was a "minimum to get things working" branch, so you can do that part.

spec/controllers/assignments_controller_spec.rb
Anbranin: I read that too. However,

> spec_helper.rb is for specs which don't depend on Rails (such as specs for classes in the lib directory)

Are you writing tests for classes in the lib directory and running them in isolation much?

> rails thinks we should have both

Well, a lot of people write rails, and yeah they have their reasons for doing this, the reason you stated being the only functional one I could discern, but I don't buy it enough. For this application, which has tests that are pretty much all about Rails, there would be no benefit. For larger projects with a lot of non-rails tests, sure. Rails was built to be flexible, and having two helper files for specs allows that. But, that flexibility also allows us to remove it.

We can get another opinion, but for now I'm gonna say if you want to have both, figure out which should go in either of them and go nuts... because I'm done with this PR. :)

spec/system/user_spec.rb
Anbranin: You're commenting on code I deleted. I totally did what you suggested.

PR #107, opened 2018-03-09
Setup system test

jforgue:


spec/system/user_spec.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'user pages' do
+  let(:membership) { create :membership, admin: true}
+  let!(:admin) { create :user, first_name: 'tester', memberships: [membership] }
```
mboneil10: Why exactly did you name the user `'tester'`?

spec/system/user_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  context 'viewing the index' do
+    it 'directs you to the appropriate page' do
+      set_current_user(admin)
```
mboneil10: Uhhhh should this sample test be involved? (AKA should I review it?)

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 100.0
+    "covered_percent": 90.18
```
mboneil10: Why the coverage drop (trying to investigate)

jforgue:
See PR #106 to better understand the changes here.

coverage/.last_run.json
mboneil10: Seems to be something happening with the helpers

mboneil10: it happened with the rails helper :(

jforgue: More specifically the view helpers aren't being tested yet all the tests pass, which shouldn't work since these are view tests. The root cause seems to be from a change in the spec_helper and/or rails_helper.

spec/system/user_spec.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'user pages' do
+  let(:membership) { create :membership, admin: true }
+  let!(:admin) { create :user, first_name: 'tester', memberships: [membership] }
```
Anbranin: This person does _not_ need a first name of "tester." That was a leftover from when I was trying different ideas for setting the current user.

spec/system/user_spec.rb
Anbranin: No, don't worry about it. This PR gets system tests set up and makes sure they work, this test should later actually be fleshed out but not right now.

spec/system/user_spec.rb
mboneil10: ok

spec/spec_helper.rb
```diff
@@ -1,17 +1,26 @@
...
-
 SimpleCov.start 'rails'
 SimpleCov.start do
   add_filter '/config/'
   add_filter '/spec/'
   refuse_coverage_drop
 end
 
+ENV['RAILS_ENV'] ||= 'test'
+require 'spec_helper'
```
Anbranin: I just noticed you're requiring the helper within itself LOL

PR #108, opened 2018-03-09
Convert Ternary to Guard Clause

mboneil10:
Perhaps a bit unnecessary

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -2,8 +2,7 @@
 
 module ApplicationHelper
   def active_class(link_path)
-    if current_page?(link_path)
-      'active'
+    return 'active' unless current_page?(link_path)
     end
   end
 end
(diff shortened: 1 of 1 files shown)


jforgue:
This is a guard clause not an if statement.

app/helpers/application_helper.rb
```diff
@@ -2,6 +2,6 @@
 
 module ApplicationHelper
   def active_class(link_path)
-    current_page?(link_path) ? 'active' : ''
+    return 'active' unless current_page?(link_path)
```
werebus: 1. I think you got this backwards. They should have the 'active' class _if_ this is the current page, not _unless_ it is.
2. Very minor nit-pick (perhaps appropriate here, ahem): You can forgo the explicit return.

So, this should be `'active' if current_page?(link_path)`

mboneil10: Thanks!

PR #109, opened 2018-04-10
Calendar access

### LINKED ISSUE #89: Calendar access
It might be nice if we could access our upcoming assignments in, e.g. Google calendar.

Two possible ways to do that:

1.  An `.ics` view for the `AssignmentsController#index`.  This needs to be excluded from Shibboleth and auth, you can use some sort of semi-secret token. ([An][1] [example][2] in a [personal][3] [project][4])
2.  Attach a calendar invite to the reminder.  (How? Dunno. MIME-shenanigans?)

[1]:  https://github.com/werebus/moretticamp/blob/master/config/routes.rb#L21
[2]:  https://github.com/werebus/moretticamp/blob/master/app/controllers/events_controller.rb#L98-L105
[3]:  https://github.com/werebus/moretticamp/blob/master/app/controllers/events_controller.rb#L27
[4]:  https://github.com/werebus/moretticamp/blob/master/app/views/events/index.ics.erb

jforgue:
Closes #89. 
All users in the database will need to be saved after this is merged.


jforgue:
Testing still needed.

app/assets/javascripts/global.js
```diff
@@ -0,0 +1,17 @@
+$(document).ready(function() {
+    var copyTextBtn = document.querySelector('.copy-text-btn');
```
Anbranin: Why did you write all of this in pure Javascript and not use JQuery?

jforgue: Almost entirely because I'm not familiar with JQuery.

Anbranin: oh no

jforgue:
Here is a screen shot with the copy button and link.
screen shot 2018-04-13 at 3 57 10 pm

app/assets/javascripts/global.js
```diff
@@ -0,0 +1,16 @@
...
+    var copyTextBtn = $('.copy-text-btn');
+
+    $(copyTextBtn).on('click', function() {
+        var copyTextarea = $('.copy-text');
+        copyTextarea.focus();
+        copyTextarea.select();
+        try {
+            var successful = document.execCommand('copy');
+            var msg = successful ? 'successful' : 'unsuccessful';
+            console.log('Copying text command was ' + msg);
```
werebus: Should we provide any feedback to the user about whether it was successful or not?

app/controllers/assignments_controller.rb
```diff
@@ -105,6 +107,18 @@ def update
     end
   end
 
+  def feed
+    user = (@current_user || User.find_by(calendar_access_token:
```
werebus: Won't `@current_user` always be `nil`? Since you skipped the before action that sets it?

app/assets/javascripts/assignment.js
```diff
@@ -1,16 +1,22 @@
...
         copyTextarea.select();
         try {
             var successful = document.execCommand('copy');
-            var msg = successful ? 'successful' : 'unsuccessful';
-            console.log('Copying text command was ' + msg);
+            console.log('Copied successfully');
         } catch (err) {
             console.log('Unable to copy');
         }
+        if(successful === true) {
```
dfaulken: Remember that boolean values are already the bread and butter of `if` statements. `successful === true` will evaluate to a boolean, but so will `successful` by itself.

jforgue: Right, can't believe I didn't realize what I did! Thanks!

jforgue: I was just too excited about threequal.

dfaulken: I mean who wouldn't be

Anbranin:
j-dawg once @werebus approves (since he commented) we can merge.

werebus:
FYI, when I comment rather than requesting changes, it means I don't care about the thing one way or the other enough to block merging.

PR #111, opened 2018-06-01
FactoryGirl -> FactoryBot

werebus:
Also did some cleanup in the spec_helper

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,47 +1,54 @@
 # frozen_string_literal: true
 
 require 'simplecov'
-SimpleCov.start 'rails'
-SimpleCov.start do
+SimpleCov.start 'rails' do
   add_filter '/config/'
   add_filter '/spec/'
   refuse_coverage_drop
 end
 
 ENV['RAILS_ENV'] ||= 'test'
-require 'spec_helper'
 require File.expand_path('../../config/environment', __FILE__)
 require 'rspec/rails'
 require 'rack_session_access/capybara'
-require 'factory_girl_rails'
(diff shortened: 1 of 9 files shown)


spec/spec_helper.rb
```diff
@@ -1,47 +1,54 @@
...
-SimpleCov.start 'rails'
-SimpleCov.start do
+SimpleCov.start 'rails' do
   add_filter '/config/'
   add_filter '/spec/'
   refuse_coverage_drop
 end
 
 ENV['RAILS_ENV'] ||= 'test'
-require 'spec_helper'
```
Anbranin: lol

spec/spec_helper.rb
```diff
@@ -1,47 +1,54 @@
...
+                 else
+                   raise ArgumentError
+                 end
+  if defined? page #Capybara
+    page.set_rack_session user_id: current_user.id
+  else #Request specs
+    session[:user_id] = current_user.id
+  end
 end
+alias set_current_user when_current_user_is
```
jforgue: Is there any reason to keep the `set_current_user` alias?

werebus: Eh, I kinda liked the way it sounded grammar-wise in some of the places I found it?

jforgue: I suppose it doesn't hurt to keep both, but it could cause minor confusion for new developers down the road.

werebus: I think it's fine. The use of `alias` makes it (fairly) clear that these are the same method. Rather than before where you had to use one or the other depending on which type of test you were writing.

PR #117, opened 2019-01-08
Bump rails

werebus:
Started as a fix for a CVE in activejob, not too hard an upgrade, though.


Gemfile
```diff
@@ -10,7 +10,7 @@ gem 'haml-rails'
 gem 'jquery-rails'
 gem 'jquery-ui-rails'
 gem 'mysql2'
-gem 'paper_trail'
+gem 'paper_trail', '~> 9.2'
```
werebus: Without updating paper_trail at all, Rails ends up limited to 5.1.x, but paper trail 10 had more breaking changes than I felt like going through right now.

Anbranin: How can that be, if the Gemfile.lock has all the parts of rails at 5.2? Is that not how this works? D: I thought that if the gem `rails` was at 5.1.x, then something like `ActionMailer` would be the same.

werebus: >  I thought that if the gem rails was at 5.1.x, then something like ActionMailer would be the same.

Yes, that's how it works. Maybe there's a misunderstanding? The `Gemfile.lock` prior to this PR had:

```
paper_trail (8.1.2)
      activerecord (>= 4.2, < 5.2)
```

Which means that simply typing `bundle update rails` wouldn't get us Rails 5.2 because it conflicts with paper_trail 8.1.2.  So, it was necessary to also upgrade paper_trail. I added the constraint, however, because I wasn't willing to dive into paper_trail 10 at this point.

Anbranin: Oh, I totally misunderstood what you were saying. I thought that you meant that in order to get Rails 5.2 going, you had to use paper_trail 10, the notion of which some Googling disavowed me. So I thought either I didn't get something or that you had made a mistake.
Cool

PR #121, opened 2019-05-16
Upgraded ruby to 2.6.1

werebus:
Also, ran/tweaked rubocop

Addresses umts/transit-it#922


.travis.yml
```diff
@@ -1,5 +1,4 @@
 language: ruby
-rvm: 2.4.1
```
Anbranin: We don't need this?

Anbranin: > If the ruby version is not specified by the rvm key, Travis CI uses the version specified in the .ruby-version file in the root of the repository if one is available.

Nevermiiiind :)

PR #122, opened 2019-06-04
Parameterize Calendar URL

jforgue:
Passes all tests, no change in coverage.
`titleize` doesn't exactly reverse `parameterize` but since names aren't case sensitive it should be OK.

### DIFF
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -109,7 +109,7 @@ class AssignmentsController < ApplicationController
 
   def feed
     user = User.find_by(calendar_access_token: params[:token])
-    roster = Roster.find_by(name: params[:roster].titleize)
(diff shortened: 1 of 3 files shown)


werebus:
> I had a hard time with "find_by isn't case-sensitive"

Likewise. It depends on the database collation. Practically speaking, we're using a (weird) `ci` collation, but it might be worth an explicit comparison to "lower" here. Especially since the _one_ roster we have in production actually has a name that gets munged by parameterizing and then titleizing:

```ruby
Roster.first.name.tap{|n| puts n.inspect}.parameterize.titleize
#=> "Transit IT"
#=> "Transit It"
```
The "problem" with something like, `.where('lower(name) = ?', name.downcase).first` is that it breaks your ability to use indexes, but in this case, it's unlikely that that the number of rosters is going to a point where that performance impact will be noticeable.

werebus:
Whoops, but you still need to titleize the param.

PR #123, opened 2019-06-21
Parameterize calendar url

jforgue:


app/controllers/assignments_controller.rb
```diff
@@ -109,7 +109,10 @@ def update
 
   def feed
     user = User.find_by(calendar_access_token: params[:token])
-    roster = Roster.where('lower(name) = ?', params[:roster].downcase).first
+    roster = Roster.where('lower(name) = ?',
```
werebus: This indentation is a little weird. Maybe it would be better to put `params[:roster].titleize.downcase` into an intermediate variable?

jforgue:
Travis is fixed in #124

PR #124, opened 2019-07-16
Bundle update and headless system tests

jforgue:


jforgue:
Apparently a travis update changed the way `xvfb` needs to be started, so that's fixed in this PR too.

spec/spec_helper.rb
```diff
@@ -30,6 +30,10 @@
   config.before :all do
     FactoryBot.reload
   end
+
+  config.before(:each, type: :system) do
```
Anbranin: Why not use `:rack_test` for those tests that don't require javascript? It's faster.

jforgue: Because we don't specify which tests use javascript, any system could have javascript which headless chrome can handle and rack test can not.

Anbranin: > we don't specify which tests use javascript

We could, it's pretty easy. And then do something like [this](https://github.com/umts/umaps-rails/blob/master/spec/rails_helper.rb#L67-L71).

PR #133, opened 2019-12-18
Travis insists that you dont have to

werebus:
That it should "just work" if your gemfile has the "Bundled With" line.
With the release of RubyGems 3.1, `gem update --system` also tries to
install bundler and waits for your OK before overwriting the gem binary.


werebus:
Indeed, the build passes without futzing with rubygems or bundler.

PR #134, opened 2020-01-02
Upgrade screaming-dinosaur to Ruby 2.6.5 and Gems

jforgue:
The biggest change and the only issue I had was with Sprockets 4. This required a new `manifest.js` in the assets directory. You can read more about that [here](https://github.com/rails/sprockets/blob/master/UPGRADING.md#manifestjs).


.ruby-version
```diff
@@ -1 +1 @@
-2.6.1
```
werebus: Not to be a wet blanket, but because of our non-ideal, lets-run-all-of-our-apps-on-one-server, setup, you will also need to simultaneously upgrade Ruby on `app3` as well as in:

* bojangles
* fleetfocus-api
* ft-forms
* jobapps
* stop-project

All 6 changes should be merged fairly close together. I'd recommend doing this in separate ruby-version-bumping-only PRs and creating a [umts/transit-it](/umts/transit-it/issues) issue to track that process.

jforgue: Ugh I'm on it

werebus: I can pre-install 2.6.5 on `app3` for you; then switching is (somewhat) easy.

jforgue: That would be nice, thank you. I created a [project](https://github.com/orgs/umts/projects/2) to track our progress.

PR #146, opened 2020-07-16
Deliver mail

jforgue:


yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
werebus, line 1: So, this is an empty file.  And that is because this project doesn't use yarn.

node_modules/.yarn-integrity
```diff
@@ -0,0 +1,10 @@
+{
```
werebus, line 1: On projects that use yarn, `/node_modules` is typically git-ignored.  And at least [some people](https://github.com/github/gitignore/blob/218a941be92679ce67d0484547e3e142b2f5f6f0/Rails.gitignore#L61-L64) think that "integrity" files should be specifically ignored too.

PR #149, opened 2020-12-23
Toggle on Rails 5.2 defaults

werebus:
Also removed the 5.1 initializer, since we are alreadiny loading 5.1 defaults in `application.rb`


jforgue (changes requested):
After some reading ([here](https://stackoverflow.com/a/50893375/8120300) and [here](https://stackoverflow.com/a/51275459/8120300)) it seems to me that we could just load `5.2` in the `application.rb` and remove all instances of `config/initializers/new_framework_defaults_*.rb`, unless we require settings from both sets of initializers. In which case we should probably add those specific settings to `application.rb`.

werebus:
Yes, that's the final step of an upgrade. This file _existing_ is a reminder of what has recently changed in case something breaks in prod.

After this PR has been deployed for a week or two, we'll remove the framework defaults files and update the `application.rb` version value.

jforgue (approved):
Looks good

PR #153, opened 2021-01-21
Remove cofeescript dependency

werebus:
It seems Rails is moving away from it as a default, and this app has exactly 5 lines of coffeescipt in it.


Anbranin (approved):
yesss

PR #158, opened 2021-02-12
Integrate the fullcalendar library

werebus:
Rather than render our own calendar table.

I think this ends up being a bit simpler. It also will make a Bootstrap upgrade [_much_ easier](https://fullcalendar.io/docs/bootstrap-theme).


Anbranin (approved):

spec/system/assignment/edit_spec.rb
```diff
@@ -95,15 +75,10 @@
...
-      visit roster_assignments_url(roster, date: date_today)
       visit edit_roster_assignment_url(roster, assignment)
       select(@last_name, from: 'User')
       click_button 'Save'
-      # Since the assignment is 7 days long
-      expect(page).to have_link @last_name,
-                                href: edit_roster_assignment_url(roster,
-                                                                 assignment),
-                                count: 7
+      expect(assignment.reload.user).to eq @new_user
```
jforgue, line 81: Considering this is a system test, wouldn't it be better confirm this by looking for something a user could see on the page?

spec/system/assignment/ics_views_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+  end
+
+  describe 'viewing the ics formatted index' do
+    let :submit do
+      set_current_user(user)
+      visit roster_assignments_path(roster, format: 'ics')
+    end
+
+    include_examples 'ics assignments feed'
```
jforgue, line 46: This is pretty neat, I didn't know you could share examples like this.

spec/system/assignment/index_spec.rb
```diff
@@ -54,98 +49,35 @@
...
-    context 'start of assignment' do
-      it 'has a left radius' do
-        create :assignment, start_date: 3.days.ago, end_date: 3.days.since,
-                            user: user, roster: roster
+    context 'clicking on an empty day' do
+      it 'sends you to create a new assignment' do
         visit roster_assignments_url(roster)
-        expect(page).to have_selector('td .cal-event.assignment-start',
-                                      count: 1)
+        execute_script 'calendar.gotoDate("2021-02-01")'
```
jforgue, line 61: Similar to my previous comment, it might be a better approach to initiate this script execution with a click.

jforgue (commented):
I have a few questions, mainly regarding your tests. It's my understanding that system tests should imitate what a user would do and see. It seems to me like some of yours don't strictly follow that idea. With that said I'm not strongly opposed if you disagree.

spec/system/assignment/index_spec.rb
werebus: Hmmm. I guess the way a "real person" would approach the task, "go to February, 2021" would be to:

1. visit the page
1. determine whether that month is in the past, now, or the future
2. if now, stop. if past, click "prev". if future, click "next".
3. keep clicking until the page has the text, "February, 2021"

Maybe it would be better just to click on the 14th of "this month", whatever that may be? I just hate introducing dependencies on "now" into tests, always seems to bite us.

.rubocop.yml
```diff
@@ -1,7 +1,10 @@
 require: umts-custom-cops
 inherit_from: .rubocop-no-umts-cops.yml
+
 # Hold on there cowpoke! Don't write them rules here!
 # Write them in .rubocop-no-umts-cops.yml!
 # Otherwise any setting here will be respected by your local Rubocop,
 # but not by CodeClimate.
 # This is because CodeClimate is not capable of running our custom cops.
+UmtsCustomCops/BeMatcherForNonDuplicableTypes:
+  Enabled: false
```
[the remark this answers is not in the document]
werebus: No, it specifically shouldn't.

This cop is only availabe to disable if we are using umts-custom-cops. If I were to add this rule to `.rubocop-no-umts-cops.yml`, then rubocop on Travis would produce an error because the cop doesn't exist.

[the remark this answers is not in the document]
werebus: BTW, see umts/custom-cops#9

spec/system/assignment/index_spec.rb
werebus: But, eh. I do it elsewhere in the spec....

spec/system/assignment/edit_spec.rb
werebus: Pros and cons, I guess. It's going to be a lot of hoop-jumping to verify this change through the browser because of the difficulty I described elsewhere in navigating to the "right" month. I could verify it through the `#show` page if it existed, but it doesn't. As it stands, verifying it through the `#index` means 30-40 clicks on the "previous" button. Of course, there [are ways](https://stackoverflow.com/a/49504778) to get Timecop's frozen time into the browser...

Incidentally, while looking into your review comments I discovered a falsely passing test. [This line][1] passes, but only because no JS means no calendar at all.

[1]: https://github.com/umts/screaming-dinosaur/pull/158/files/aa749c6ab402e54c324d869084e47eebb2c97392#diff-a9decbd70a26c0af70dd7eac12b55d4ef79448c3a8937cc446eecad4b9d5df68R60

spec/system/assignment/index_spec.rb
jforgue: I think it's probably fine to not click as a user if you think it's trivial and it saves on complexity and test time. I'll leave the decision to you.

werebus:
@Anbranin, do you want to weigh in on @jforgue's though  that verification of changes should also be done via the browser? It can be done, with a little work; I can think of four possible ways:

1. Add a `#show` page for assignments; look there.
2. Add a JS library for time-mocking, populate that with Timecop's time
3. Add a UI to the page for direct date navigating — enter a date, click go
4. Write a spec helper that clicks the next/prev buttons as needed until the right month is in view.

jforgue:
If you are going to go through the effort, this is the option I'm leaning towards.

> 2. Add a JS library for time-mocking, populate that with Timecop's time

jforgue:
Adding a JS library for time-mocking could be useful for future tests, won't slow down our tests as much as clicking through, and doesn't add unnecessary features to our site.

app/assets/stylesheets/assignments/calendar.scss
```diff
@@ -1,67 +1,19 @@
 #calendar {
-  border-collapse: collapse;
-  border-spacing: 0;
-  width: 98%;
-  height: 100%;
-  min-width: 800px;
-  margin: 0% 1%;
```
jforgue, line 7: I would still like some margin on the sides and bottom of the calendar. I don't like that the calendar ends where the color of the main-content ends on the sides and there should be a bit more space between the bottom and the calendar link.

Screenshot 2021-02-17 131707

Anbranin:
>  do you want to weigh in on @jforgue's though that verification of changes should also be done via the browser?

I disagree that verification of changes should _always_ be done in the browser. As long as there are other tests that verify that "when someone owns a calendar event, they show up thusly" we're all set, because here you've made an assertion about the underlying database object. You've checked thoroughly enough, in other words.

PR #161, opened 2021-02-19
Add unique indeces where called for

werebus:
I was working on rubocopping the app a bit and the rails cop recommended these indeces. They're a good idea, but thought it warranted a separate PR


jforgue (commented):
Looks good to me

jforgue (approved):

PR #163, opened 2021-02-19
Name the ics feed route

werebus:
This way we don't have to construct URLs with string-interpolation and have hard-coded domains.


jforgue (approved):
Makes sense to me

PR #164, opened 2021-02-22
Update bootstrap to 4.6

werebus:
I'd recommend checking this one out and clicking around while reviewing.  This one is heavy on the visual changes.


app/views/users/_form.html.haml
```diff
@@ -0,0 +1,43 @@
...
+           options_from_collection_for_select(Roster.all, :id, :name,
+            selected: selected_rosters(@user, @roster)), {}, multiple: true, class: 'form-control'
+        %div
+          = f.check_box :reminders_enabled
+          = f.label :reminders_enabled,
+           'Receive email reminders of upcoming assignments?'
+        %div
+          = f.check_box :change_notifications_enabled
+          = f.label :change_notifications_enabled,
+           'Receive notifications when your assignments are changed?'
```
jforgue, line 35: Same as above.
```suggestion
           'Receive notifications when your assignments are changed'
```

app/views/users/_form.html.haml
```diff
@@ -0,0 +1,43 @@
...
+        - if @current_user.admin_in? @roster
+          %div
+            = f.fields_for @user.membership_in(@roster) || :membership do |g|
+              %div
+                = g.check_box :admin
+                = g.label :admin, "Admin in #{@roster.name}?"
```
jforgue, line 41: I know you didn't add this functionality, but we probably shouldn't allow the only admin of a roster to make themselves not the admin of said roster.

jforgue (changes requested):
The styling all looks good to me, just a few issues and suggestions.

app/views/users/_form.html.haml
werebus: #170

jforgue (approved):

app/helpers/application_helper.rb
```diff
@@ -1,8 +1,17 @@
 # frozen_string_literal: true
 
 module ApplicationHelper
-  def active_class(link_path)
-    'active' if current_page?(link_path)
+  def calendar_instructions
+    'Use this address to subscribe to this calendar in ' \
+    'another application (i.e. Google Calendar).'
+  end
```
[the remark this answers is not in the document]
werebus: So other options I can think of are:

* A global constant
* Local variable set in the view
* Instance variable set in the controller

I honestly like this the best of those 4 options. The purpose of helpers is to abstract view/display logic that doesn't go in a model (doesn't have to do with your app's persisted objects) and doesn't go in controllers (doesn't have to do with managing HTTP requests and responses).

PR #171, opened 2021-03-18
Stop an admin from removing the only roster admin

werebus:
* Validation on `Membership`
* Errors in the flash in the controller
* Disable the checkbox too

Closes #170

PR #172, opened 2021-03-18
Indicate admins on dev login

werebus:
This would have been nice when working on #171

### DIFF
--- a/spec/models/roster_spec.rb
+++ b/spec/models/roster_spec.rb
@@ -156,12 +156,12 @@ RSpec.describe Roster do
 
     it 'has admins in the "Admins"' do
       expect(call.fetch('Admins'))
-        .to contain_exactly(*admins.map { |a| [a.full_name, a.id] })
+        .to match_array(admins.map { |a| [a.full_name, a.id] })
     end
 
     it 'has non-admins in the "Non-Admins"' do
       expect(call.fetch('Non-Admins'))
-        .to contain_exactly(*non_admins.map { |na| [na.full_name, na.id] })
+        .to match_array(non_admins.map { |na| [na.full_name, na.id] })
     end
   end
 end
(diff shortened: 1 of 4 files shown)


spec/models/roster_spec.rb
```diff
@@ -147,4 +147,21 @@
...
+
+  describe 'user_options' do
+    let(:roster) { create :roster }
+    let!(:admins) { [roster_admin(roster)] }
+    let!(:non_admins) { [roster_user(roster)] }
+    let(:call) { roster.user_options }
+
+    it 'has admins in the "Admins"' do
+      expect(call.fetch('Admins'))
+        .to contain_exactly(*admins.map { |a| [a.full_name, a.id] })
```
jforgue, line 159: Wouldn't `expect(call.fetch('Admins')).to eq(admins.map { |a| [a.full_name, a.id] })` be sufficient?

jforgue (approved):
Looks good.

Anbranin (approved):

spec/models/roster_spec.rb
jforgue: or maybe `match_array` would be the more appropriate matcher?

Anbranin: What's wrong with `contain_exactly`? I use it all the time so I'm curious

werebus: There are two different improvement paths @jforgue proposed here. In the first, he was pointing out that, in this test, there is only one admin, and a one-element array can only be in one order, so `eq` would work here. However, I prefer to explicitly say "in no particular order", I think.

In the second, he suggests `match_array` which would get rid of the need for the splat.  That is, these two expectations are the same:

```ruby
expect(a_thing).to contain_exactly(:a, :b, :c)
expect(a_thing).to match_array([:a, :b, :c])
```

PR #173, opened 2021-03-24
Finish Rails 6 - load defaults

werebus:
We've been running all 6.0 defaults since 3/5


jforgue (approved):
> We've been running all 6.0 defaults since 3/5

Neat; looks good to me.

PR #177, opened 2021-04-22
Rails 6.1 upgrade

werebus:


bin/spring
```diff
@@ -1,15 +0,0 @@
...
-unless defined?(Spring)
-  require "rubygems"
-  require "bundler"
-
-  if match = Bundler.default_lockfile.read.match(/^GEM$.*?^    (?:  )*spring \((.*?)\)$.*?^$/m)
-    Gem.paths = { "GEM_PATH" => [Bundler.bundle_path.to_s, *Gem.path].uniq }
-    gem "spring", match[1]
-    require "spring/binstub"
-  end
-end
```
Anbranin, line 15: Why delete this file? As far as I can tell the file should exist in Rails 6.1, it will just be different.

werebus: The app task deleted it [because Spring is not in our Gemfile, and hence undefined](https://github.com/rails/rails/blob/8e7c3f407a7b5f37d2e8fd63110cf2fbff135f94/railties/lib/rails/app_updater.rb#L32).

We could keep it, but the install process (`spring binstub --all`) will generate it again if we ever do decide to install Spring on this app.

Anbranin (approved):

PR #185, opened 2021-05-04
Fix rotation generator

werebus:
Turns out that that submit button wasn't in the form. Ugh.


jforgue (approved):
Sometimes end tags really are nice. Still better than erb though.

Anbranin (approved):

PR #193, opened 2021-05-18
Update umts-custom-matchers

werebus:
(In order to update rspec-rails)


Gemfile.lock
```diff
@@ -326,8 +326,8 @@ GEM
       concurrent-ruby (~> 1.0)
     uglifier (4.2.0)
       execjs (>= 0.3.0, < 3)
-    umts-custom-matchers (1.0.0)
-      rspec-rails (>= 3.0, <= 5.0)
```
werebus, line 330: Huh, this was `<=`?  That seems like a mistake; oh well.

Anbranin: Wait, why does it seem like a mistake?

Anbranin (approved):

Gemfile.lock
werebus: It's a version mistake in [`umts-custom-matchers`](https://github.com/umts/custom-matchers/commit/a015a604741784dc23274f75d9d267d00d8f6af0#diff-f30118d2c6b519f807837fa1eef26f495d35289c45a5a001d7da79028493f4ffR27), in my opinion. That version was carried forward into the new version too.

This constraint allows, by example, the following versions: 3.0.0, 3.22.99, 4.0.7, 4.3.12, 5.0.0.  But not 5.0.0.1

PR #221, opened 2022-10-26
Update `bundler`

bandogora:
Do we nee to manually install the new version on the server?


benmelz (approved):
> Do we nee to manually install the new version on the server?

I think we do in order for the server to actually _use_ the new bundler version, that being said I also think that it won't fail because of a bundler version mismatch (at least it usually doesn't locally).

PR #227, opened 2022-11-08
Two fixes for compiling assets in production

werebus:
1. Rails 7 no longer automatically runs `yarn` when precompiling assets. Wasn't mentioned in the release notes, but see rails/rails#43641. Hello again, my friend, `capistrano-yarn`.
2. Uglifier was failing on a `const`... somewhere. I really have no idea where. Uglifier themselves say to use `terser` if you want ES6 support, so...

### DIFF
--- a/Capfile
+++ b/Capfile
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-%w[setup deploy scm/git pending bundler rails passenger]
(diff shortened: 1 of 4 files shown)


bandogora (approved):
This looks reasonable for now. I would like to switch over to import maps in place of yarn in the future.

PR #228, opened 2022-11-08
Don't call each on `flash[:errors]` unless it's an array

### LINKED ISSUE #226: Calling `each` on `flash[:errors]` returns `undefined method 'each' for :String`
https://github.com/umts/screaming-dinosaur/blob/24b5b2bd8cb0033447457fa0443e1faf4a83497d/app/views/layouts/application.haml#L22-L23

As far as I can tell we only test that`flash[:errors]` isn't empty, not that they're displayed correctly.

bandogora:
- Only list errors if `flash[:errors]` is an Array
- Test generate rotation starting user not in rotation

Closes #226


app/views/layouts/application.haml
```diff
@@ -11,14 +11,16 @@
...
-        .alert.alert-success
+        .alert.alert-success{ role: 'alert' }
           = flash[:message]
           - if flash[:change].present?
             (#{link_to 'undo', undo_change_path(flash[:change])})
       - if flash[:errors].present?
-        .alert.alert-danger
-          The following errors occurred:
-          %ul
-            - flash[:errors].each do |error_message|
```
dirschn, line 22: How about just `Array(flash[:errors]).each`? it won't do anything if it's already an array, but will then work if it isn't.

bandogora: I don't want to create a list for a single message.

app/controllers/assignments_controller.rb
```diff
@@ -143,11 +143,9 @@ def render_ics_feed
   def require_taking_ownership
     return true if @current_user.admin_in?(@roster) || taking_ownership?
 
-    flash[:errors] = [<<-TEXT]
-      You may only edit or create assignments such that you become on call.
-      The intended new owner of this assignment must take it themselves.
-      Or, a roster administrator can perform this change for you.
-    TEXT
+    flash[:errors] = 'You may only edit or create assignments such that you become on call. ' \
```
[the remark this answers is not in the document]
bandogora: It's a bit unusual compared to how we normally format error messages.

app/views/layouts/application.haml
dirschn: Okay, I usually try to cut down on lines but I suppose this method sort of cuts down on complexity.

spec/system/assignment/edit_spec.rb
```diff
@@ -68,9 +68,10 @@
     it 'stops them from changing assignments they do not own' do
       select(new_user.last_name, from: 'User')
       click_button 'Save'
-      within('div.alert.alert-danger') do
-        expect(page).to have_selector 'li', text: 'You may only edit'
-      end
+      expect(page).to have_selector '.alert.alert-danger',
+                                    text: 'You may only edit or create assignments such that you become on call. ' \
```
dirschn, line 72: Are you able to use the localized format here? might as well, right? Unless i suppose you reaalllyyy want it to be very explicit

bandogora: This isn't in a controller so we don't have access to the locales.

dirschn (approved):

PR #229, opened 2022-11-09
Allow only roster admins to delete assignments

### LINKED ISSUE #65: Should anyone be able to destroy any assignment?
In a certain sense, being on call is equivalent to money. Should anyone be able to get rid of anyone else's potential future money?

Our current system is built around the idea of accepting responsibility, but it's worth noting this also.

bandogora:
Closes #65


app/controllers/assignments_controller.rb
```diff
@@ -25,10 +25,15 @@ def create
...
-    @assignment.destroy
-    confirm_change(@assignment)
-    redirect_to roster_assignments_path(@roster)
+    if @current_user.admin_in?(@roster)
+      @assignment.notify :owner, of: :deleted_assignment, by: @current_user
+      @assignment.destroy
+      confirm_change(@assignment)
+      redirect_to roster_assignments_path(@roster)
+    else
+      flash[:errors] = 'Only roster admins may delete assignments.'
```
dirschn, line 34: I know that @benmelz  has expressed the opinion of wanting to start using locale based text in round-three, I would assume that should apply here as well?

bandogora: If rubocop was updated I would have seen this!

dirschn (approved):
seems fine to me

PR #232, opened 2022-11-14
Update linters and rearrange `Gemfile` groups

bandogora:
I pretty much just copied over `.haml_lint.yml` and `.rubocop.yml` from round-three and added the `rubocop-performance` gem. I also moved some gems to more appropriate bundler groups in the `Gemfile`.

Edit: Almost forgot `eslint`.


.codeclimate.yml
```diff
@@ -14,6 +14,7 @@ plugins:
     enabled: true
   rubocop:
     enabled: true
+    channel: rubocop-1-31-0
```
dirschn, line 17: image
From https://docs.codeclimate.com/docs/rubocop

Do you plan on adding one?

bandogora: We already have a `.rubocop.yml` that's compatible with that version

dirschn: Okay cool just checking

.rubocop.yml
```diff
@@ -1,42 +1,84 @@
...
 
 AllCops:
   Exclude:
+    - './*'
     - 'bin/**/*'
-    - 'config/{environments,initializers}/**/*'
-    - 'config/{application,boot,environment,puma,spring}.rb'
+    - 'config/**/*'
     - 'db/**/*'
     - 'script/**/*'
```
dirschn, line 12: There is no `script/` directory in this app

dirschn (approved):
this seems fine to me

PR #235, opened 2022-11-15
Replace `ab-datepicker` with browser datepicker

bandogora:
we definitely don't need this now that browser HTML date pickers are good.


dirschn (approved):
seems fine to me

PR #239, opened 2022-11-18
Only allow admins to select different users for assignments

### LINKED ISSUE #220: Default to current user when creating a new assignment
For example, if you go to https://umts-oncall.admin.umass.edu/rosters/1/assignments/new?date=2023-01-27 it selects the user at the top of the list. A non-admin user can't create new assignments for other users, so it would probably be cleaner if it defaulted to the current user.

bandogora:
Closes #220


bandogora:
> Was the idea that we can't give other people our assignments, we can only take them from others?

That is correct, unless you are an admin in that roster.

dirschn (commented):
It seems to me that you'd want to add more robust guards, specifically within the controller action to guard for this? I know that this is good _enough_ in most cases, but nothing is preventing someone from a little browser manipulation and then submitting the "wrong" user.

I know we are a standup group, but I still think frontend restrictions should be mirrored in the backend.

bandogora:
> It seems to me that you'd want to add more robust guards, specifically within the controller action to guard for this? I know that this is good _enough_ in most cases, but nothing is preventing someone from a little browser manipulation and then submitting the "wrong" user.

The gaurds are already in place:
https://github.com/umts/screaming-dinosaur/blob/a57901163f462e351a0f51b0a9c0f398d9f13d0e/app/controllers/assignments_controller.rb#L17
https://github.com/umts/screaming-dinosaur/blob/a57901163f462e351a0f51b0a9c0f398d9f13d0e/app/controllers/assignments_controller.rb#L80

dirschn:
Ah, I misunderstood what this PR was supposed to do. I get it now.

app/views/assignments/_form.html.haml
```diff
@@ -4,8 +4,13 @@
...
         = f.label :user_id
-        = f.collection_select :user_id, @users, :id, :last_name,
-                              {}, class: 'form-control custom-select'
+        - if @current_user.admin_in?(@roster)
+          = f.collection_select :user_id, @users, :id, :last_name,
+            { selected: @assignment.user&.id || @current_user.id }, class: 'form-control custom-select'
+        - else
+          = f.collection_select :user_id, @users, :id, :last_name,
+            { disabled: @users.reject { |u| u == @current_user }.pluck(:id), selected: @current_user.id},
+            class: 'form-control custom-select'
```
werebus, line 13: I don't think there's much point to a select box that will only ever have one option you can select

```suggestion
          = f.hidden_field :user_id, value: @current_user.id
          = f.text_field :user_name, value: @current_user.last_name, disabled: true, class: 'form-control'
```

werebus (changes requested):

werebus (approved):

PR #240, opened 2022-11-21
Use `bootstrap-icons` instead of `fontawesome-free`

bandogora:
This was a pretty easy conversion, and since the Full Calendar `bootstrap5` theme uses `bootstrap-icons` by default I figured it was worth while.


benmelz:
This might warrant a group conversation, especially if this is gonna bubble out to our other apps. I don't have a decided opinion yet, but two red flags worth considering might be:

- icon library size (at one point in time `fontawesome` had a much larger free icon set, although that may not be the case anymore)
- support (we've been burned by `glyphicons` disappearing in the past, how sure are we that this iteration is gonna stick around?)

PR #241, opened 2022-11-21
Add `scss_lint` and fix `haml-lint`

bandogora:
Adds `scss_lint` to match code climate and fixes the `haml-lint` config file name.


benmelz (changes requested):
https://github.com/umts/round-three/pull/5395#pullrequestreview-1188856081

PR #244, opened 2022-11-23
Check login user for subsidiary account

### LINKED ISSUE #205: Subsidiary accounts should not be allowed to log in
Microsoft's Auth has me logged in as my `af` account and when I went to umts-oncall.admin.umass.edu I was logded in as my supervisor @sherson.

bandogora:
Closes #205.

I tried to right a request spec for this, but it turns out setting a `request.env` var is a pain. You can pass `env: { uid: 12 }` to the `get` method, but it merges that env var into the Rack env hash. That auto-prefixes it so it ends up as `HTTP_UID`.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,8 +3,7 @@
 class ApplicationController < ActionController::Base
   attr_accessor :current_user
 
-  before_action :check_primary_account
-  before_action :set_current_user, :set_roster, :set_paper_trail_whodunnit
+  before_action :check_primary_account, :set_current_user, :set_roster, :set_paper_trail_whodunnit
 
   def confirm_change(object, message = nil)
     change = object.versions.where(whodunnit: @current_user).last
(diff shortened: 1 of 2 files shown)


app/views/sessions/subsidiary.html.haml
```diff
@@ -0,0 +1,11 @@
+%h1 Welcome to the website for Special Transportation.
```
werebus, line 1: Wrong site

app/controllers/application_controller.rb
```diff
@@ -3,6 +3,7 @@
 class ApplicationController < ActionController::Base
   attr_accessor :current_user
 
+  before_action :check_primary_account
   before_action :set_current_user, :set_roster, :set_paper_trail_whodunnit
```
werebus, line 7: As evidenced by the second line, these can be combined.

werebus (changes requested):

werebus (approved):

PR #247, opened 2022-11-29
Bootstrap 5

bandogora:
I did make a few non-essential changes to make our forms and spacing consistent. Should be pretty straight forward though.

https://getbootstrap.com/docs/5.0/migration/


app/views/layouts/_nav.haml
```diff
@@ -17,8 +16,6 @@
               rotation_generator_roster_assignments_path(@roster)
       %ul.navbar-nav
         - if @current_user.present?
-          %li.nav-item.navbar-text.mx-2
-            Logged in as #{@current_user.full_name}
-          = nav_link_item 'Edit Profile',
+          = nav_link_item @current_user.full_name,
```
bandogora, line 19: Bootstrap 5 changed uses your browsers font by default now. I noticed the nav-bar got pretty squished because of that. I feel it's a pretty normal behavior to expect your name in a nav-bar to bring you to your profile, so I opted to make that change.

app/assets/javascripts/assignment.js
```diff
@@ -1,8 +1,11 @@
 $(document).ready(function() {
-  // enable bootstrap tooltips
-  $("[data-toggle='tooltip']").tooltip('enable');
+  // Enable bootstrap tooltips
+  var tooltipTriggerList = document.querySelectorAll('[data-bs-toggle="tooltip"]');
+  tooltipTriggerList.forEach(function(tooltipTriggerNode) {
+    new bootstrap.Tooltip(tooltipTriggerNode);
+  });
```
benmelz, line 6: I'm still questioning why we need to assign `var tooltipTriggerList`

app/assets/javascripts/application.js
```diff
@@ -12,7 +12,7 @@
 //
 //= require jquery
 //= require jquery-ujs
-//= require popper.js/
+//= require @popperjs/core/dist/umd/popper
```
benmelz, line 15:
```suggestion
//= require @popperjs/core
```

I think the default cjs version is fine to import

app/views/assignments/_current.haml
```diff
@@ -1,11 +1,13 @@
...
+      .card-body.bg-success
+        %h5.card-title Currently on call: #{@current_assignment.user.last_name}
+    - else
+      .card-body.bg-warning
+        %h5.card-title No one is currently on call!
+        - if @fallback_user.present?
+          .card-text #{@fallback_user.last_name} is the fallback.
+  .card.mb-3
+    .card-body
+      .card-text On-call switches at: #{switchover_time(@switchover_hour)}
```
benmelz, line 13: Sidenote (no change being requested) these are both conceptually and styled pretty similarly to `alert`s.

benmelz (changes requested):
After looking at two of the files involved, the form adjustments are gonna make this a _very_ dense review. I very much urge you factor them out and pare this back to a "upgrade while preserving previous styling", else it's gonna be a long haul

app/assets/javascripts/application.js
werebus: I'm really not that good at Javascript, but:

```
Uncaught ReferenceError: exports is not defined
```

app/views/assignments/_current.haml
werebus: I've got a PR ready based on this one

benmelz (approved):

PR #251, opened 2022-12-08
Use a slightly more forgiving coverage config

werebus:
This is what we have on st-pax and incident. By restricting it to 'CI', it also allows a dev to explicitly allow _any_ coverage drop if they want to as long as they commit the change to the coverage file -- if it's an unacceptable drop, that can be caught in code review.


coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "line": 100.0
+    "line": 99.32
```
werebus, line 3: This drop was actually introduced in #244, for example.

benmelz (approved):

PR #254, opened 2022-12-29
Flip on most Rails 7.0 defaults

werebus:
Just a few of the more complicated ones left.


app/controllers/changes_controller.rb
```diff
@@ -14,6 +14,6 @@ def undo
       version.item.destroy
       flash[:message] = "#{version.item_type} has been deleted."
     end
-    redirect_back fallback_location: 'public/404.html'
+    redirect_back fallback_location: '/public/404.html'
```
werebus, line 17: This was a bug exposed by the new "open redirect" protection.

In test, at least, this was generating redirects to 'http://test.hostpublic/404.html', which is not the same host as 'test.host' and therefore now raises an exception.

config/initializers/cookies_serializer.rb
```diff
@@ -1,5 +0,0 @@
-# Be sure to restart your server when you modify this file.
-
-# Specify a serializer for the signed and encrypted cookie jars.
-# Valid options are :json, :marshal, and :hybrid.
-Rails.application.config.action_dispatch.cookies_serializer = :json
```
werebus, line 5: This is now the default (and is in the "new framework defaults" file).

config/initializers/wrap_parameters.rb
```diff
@@ -1,14 +0,0 @@
-# Be sure to restart your server when you modify this file.
-
-# This file contains settings for ActionController::ParamsWrapper which
-# is enabled by default.
-
-# Enable parameter wrapping for JSON. You can disable this by setting :format to an empty array.
-ActiveSupport.on_load(:action_controller) do
-  wrap_parameters format: [:json]
-end
```
werebus, line 9: The 7.0 default is 

```ruby
Rails.application.config.action_controller.wrap_parameters_by_default = true
```

PR #259, opened 2023-01-30
Email warning to admin if there are uncovered upcoming days

### LINKED ISSUE #219: Email warning to admin if there are uncovered upcoming days
Perhaps a periodic (daily?) check could be done for any uncovered days over next X days, and if there are any, an email could be sent to the admin (could be a useful safety check since we allow users to change days of their own assignments, which could mistakenly result in uncovered days).

bandogora:
Closes #219.


bandogora:
Sample email:
Screenshot from 2023-01-30 13-47-44

spec/models/roster_spec.rb
```diff
@@ -215,4 +215,18 @@
         .to match_array(non_admins.map { |na| [na.full_name, na.id] })
     end
   end
+
+  describe 'check_for_open_dates_between' do
```
benmelz, line 219:
```suggestion
  describe '#check_for_open_dates_between' do
```

spec/mailers/roster_mailer_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+        it 'includes the fallback user name' do
+          expect(email.body.encoded).to have_text "The fallback user is #{roster.fallback_user.full_name}."
+        end
+      end
+    end
+
+    context 'without open assignments in the next two weeks' do
+      before { create(:assignment, roster: roster, start_date: Time.zone.today, end_date: 2.weeks.from_now) }
+
+      it 'does not ques the email to be sent' do
```
benmelz, line 60:
```suggestion
      it 'does not queue the email to be sent' do
```

spec/mailers/roster_mailer_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    let(:roster) { create(:roster) }
+    let(:admin) { create(:user) }
+
+    before { create(:membership, roster: roster, user: admin, admin: true) }
+
+    context 'with open assignments in the next two weeks' do
+      before { create(:assignment, roster: roster, start_date: 1.week.from_now, end_date: 2.weeks.from_now) }
+
+      it 'ques the email to be sent' do
```
benmelz, line 15:
```suggestion
      it 'queues the email to be sent' do
```

benmelz (changes requested):
misc. typos and such

app/models/roster.rb
```diff
@@ -69,4 +69,12 @@ def user_options
     nas = non_admins.order(:last_name).map { |na| [na.full_name, na.id] }
     { 'Admins' => as, 'Non-Admins' => nas }
   end
+
+  def check_for_open_dates_between(start_date, end_date)
```
benmelz, line 73: I personally am not a fan of `check_` style naming for methods that return data, especially if no "action" is being performed in accordance with the check. IMO it's a lot more ruby-like to just name the method according to what's returned, i.e.

```suggestion
  def uncovered_dates_between(start_date, end_date)
```

Just my opinion though

config/schedule.rb
```diff
@@ -5,3 +5,7 @@
 every :day, at: '9:00am' do
   runner 'Assignment.send_reminders!'
 end
+
+every :day, at: '4:00am' do
+  runner 'Roster.find_each { |roster| RosterMailer.with(roster: roster).open_dates_alert.deliver_now }'
```
benmelz, line 10: I don't really love having anything more than a task/method call encoded in these runner strings. It brings application logic out of the application and also makes it untestable.

For these types of things I personally like writing `Job`s (say, a `CheckUncoveredDatesJob`) and calling them here like

```suggestion
  runner 'CheckUncoveredDatesJob.perform_now'
```

both making it testable and encapsulating it into a nice unit of work.

app/mailers/roster_mailer.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class RosterMailer < ApplicationMailer
+  before_action { @roster = params[:roster] }
+
+  def open_dates_alert
+    @open_dates = @roster.check_for_open_dates_between(Time.zone.now, 2.weeks.from_now)
+    mail to: @roster.admins.pluck(:email), subject: "Upcoming dates are uncovered for #{@roster.name} On-Call"
+    mail.perform_deliveries = false if @open_dates.empty?
```
benmelz, line 9: I'm not sure I'm really on board with having the `perform_deliveries = false` logic in the mailer itself here. To me it doesn't make sense that a mailer should sometimes "cancel itself" - the responsibilities of how/whether or not to send an email are usually delegated to the callee (i.e. usually callee decides whether to `deliver_now`, `deliver_later`, or not send at all).

Keeping in the vein of my review, such responsibility would fall on the `Job` (i.e. _it_ would gather the `open_dates` data for each roster, determine whether or not to send an email for, then just forward those dates here for formatting).

bandogora: I don't disagree, so I'll make the change. With that said, `mail.perform_deliveries = false` is one of the `after_action` examples in the [Action Mailer Rails guide](https://guides.rubyonrails.org/action_mailer_basics.html#action-mailer-callbacks) .

benmelz: Interesting - I gave it a quick readthrough but missed that section.

app/models/roster.rb
```diff
@@ -69,4 +69,12 @@ def user_options
     nas = non_admins.order(:last_name).map { |na| [na.full_name, na.id] }
     { 'Admins' => as, 'Non-Admins' => nas }
   end
+
+  def uncovered_dates_between(start_date, end_date)
+    open_dates = []
+    start_date.to_date.upto(end_date.to_date).each do |date|
+      open_dates << date if Assignment.where(roster: self).between(date, date).empty?
+    end
+    open_dates
```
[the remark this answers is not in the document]
werebus: You can do one better than `tap`:

```suggestion
    (start_date.to_date...end_date.to_date).select do |date|
      assignments.between(date, date).empty?
    end
```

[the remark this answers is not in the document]
benmelz: I'll also offer this up - an optimized implementation:

```suggestion
    (start_date.to_date..end_date.to_date).to_a - assignments.between(start_date, end_date)
                                                             .inject([]) do |dates, assignment|
      dates | (assignment.start_date..assignment.end_date).to_a
    end
```

I noticed that this has a query complexity of `O(n)` with respect to the number of assignments in the date range (each `.empty?` call will produce a SQL query). I couldn't resist tinkering and came up with this single-query version.

It works by collecting all of the covered dates (the `assignment.between(..).inject([])...`), then subtracting them from the full set of dates (i.e. `uncovered_dates = all_dates - covered_dates`).

In reality it doesn't really matter since this is gonna run in the background though, so do with it what you will.

app/jobs/check_rosters_uncovered_dates_job.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CheckRostersUncoveredDatesJob < ApplicationJob
+  queue_as :default
+
```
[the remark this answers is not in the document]
bandogora: This was added by the rails generator. I can remove it, but I imagine it doesn't hurt to leave it in.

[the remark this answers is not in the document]
bandogora: I suppose we don't actually have a queue adapter so it is a little misleading.

benmelz (approved):

PR #260, opened 2023-02-02
Update README.md

bandogora:
We're not on rails 5 anymore and we don't need to manually run every task from `db:setup`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,11 +12,7 @@ A Rails app for the management of on-call schedules and interactions with Twilio
 ## Development
 
 ### Setup
-1. Install Gems: `bundle`
-2. Install Node packages: `yarn`
-3. Create your application.yml: `cp config/application.yml.example config/application.yml`
-4. Create your database.yml: `cp config/database.yml.example config/database.yml`
-5. Setup your databases: `rails db:setup`
+Run `bin/setup`
 
 ### Style guides
 
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # screaming-dinosaur
 
+A Rails app for the management of on-call schedules and interactions with Twilio.
+
 ![screaming dinosaur](app/assets/images/screaming_dinosaur.jpg)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,33 +1,22 @@
...
+1. Install Gems: `bundle`
+2. Install Node packages: `yarn`
 3. Create your application.yml: `cp config/application.yml.example config/application.yml`
 4. Create your database.yml: `cp config/database.yml.example config/database.yml`
-5. Create the databases: `rails db:create`
-6. Setup databases: `rails db:schema:load`
-7. Seed development data: `rails db:seed`
-
-When seeding, you can skip creating assignments with `SKIP_ASSIGNMENTS=true rake db:reset` etc.
+5. Setup your databases: `rails db:setup`
```
benmelz, line 19: I'll raise you one and point out that this whole section can be replaced with:

```sh
bin/setup
```

(although if we're putting _that_ down maybe now's a good time to move the `setup` and `update` scripts from the `bin/` stub folder to `script/`)

bandogora: Oh neat! That's new to me.

benmelz (approved):

README.md
benmelz: If I remember correctly, the main reason would be organization - I think this was a conversation I had with @werebus a long time ago but I'm shakier on it now. As is on round-three, the classifications are "`bin/` for all of rails' binstubs specifically, `script/` for all scriptable development tasks (i.e. `setup`, `update`, `server` when using foreman, etc)".

I'd be open to revisiting that conversation I guess though.

bandogora: In case you haven't already seen it, the [Rails Getting Started Guid](https://guides.rubyonrails.org/getting_started.html#creating-the-blog-application) says: 

> [bin/] Contains the rails script that starts your app and can contain other scripts you use to set up, update, deploy, or run your application.

benmelz: Yeah I'm aware, I think back then it was just a case of us disagreeing with rails' opinion on the matter

PR #276, opened 2023-06-14
Use Rails built-in secure token impementation

werebus:
I was skimming some Rails blogs and found that Rails already has a built in solution for "user needs a random token".

I went back and forth on dealing with uniqueness here. Collisions are _extremely_ unlikely. In the event that there is one, it would just mean that one user could read another user's identical copy of the assignments feed. **But**, I could foresee some future developer assuming that that token guarantees a user's identity in some future feature.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -12,16 +12,13 @@ class User < ApplicationRecord
                               inverse_of: 'fallback_user',
                               dependent: :nullify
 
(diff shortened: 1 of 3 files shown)


benmelz:
Ah that makes more sense. I still don't understand how/when the schema gets loaded in travis, but yeah approval stands

PR #277, opened 2023-06-23
fix error case for form submission

dirschn:
Was gonna fix all the forms but there's a lot of... weird stuff going on with this site, so will either look into it next week or will be told "don't bother"

### DIFF
--- a/spec/controllers/users_controller_spec.rb
+++ b/spec/controllers/users_controller_spec.rb
@@ -39,7 +39,7 @@ RSpec.describe UsersController do
         end
 
         it 'redirects back' do
-          expect { submit }.to redirect_back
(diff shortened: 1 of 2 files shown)


benmelz (approved):

benmelz (changes requested):
Oops spoke too soon - tests need to be updated

spec/controllers/users_controller_spec.rb
```diff
@@ -39,7 +39,7 @@
         end
 
         it 'redirects back' do
```
benmelz, line 41: Should probably be updated, as none of these `'redirect back'` anymore

benmelz (changes requested):

benmelz (approved):

PR #287, opened 2023-07-20
Fix test failure after 8pm

werebus:
My _browser_ is in the EDT time zone. But in UTC it's already tomorrow. So `Time.zone.today.day` returns a different number than the calendar on the site.

Failing test locally:

https://github.com/umts/screaming-dinosaur/blob/1ba508a7fb5d812626a2ab11c5e1b36444bbf0fd/spec/system/assignment/index_spec.rb#L39-L43


werebus:
Won't be able to see if this works until #288 is fixed.

benmelz (approved):

Issue #288, opened 2023-07-20
Webdrivers broken on Chrome 115+


werebus:
(Which is "latest" on Travis)

See: titusfortner/webdrivers#247

TLDR, beginning with Chrome 115, Google changed how it publishes chromedriver versions, and the old download location is deprecated.

Possible fixes:
* Pin chromedriver to 114 - this will work until Chrome 116 is released, as chromedriver will work up to one version ahead
* Upgrade `selenium-webdrivers` to 4.11, that gem version has a new snazzy chromedriver version manager built into it that will work with the new location. Unfortunately, 4.11 isn't released yet.

PR #289, opened 2023-07-20
Temporarily pin our `chromedriver` version to 114.x

werebus:
Bandaid for #288

It's a race. Hopefully, `selenium-webdriver` 4.11 comes out before Chrome 116.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -18,8 +18,8 @@ Capybara.enable_aria_label = true
 
 # Temporary fix for change in chromedriver 115 download location.
 # After `selenium-webdriver` 4.11 is released, we can use just that
-# instead of `webdrivers`.
-Webdrivers::Chromedriver.required_version = "114.0.5735.90"
+# instead of `webdrivers`. This will break when Chrome 116 is released.
+Webdrivers::Chromedriver.required_version = '114.0.5735.90'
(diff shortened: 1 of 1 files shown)


spec/spec_helper.rb
```diff
@@ -16,6 +16,11 @@
 Capybara.server = :puma, { Silent: true }
 Capybara.enable_aria_label = true
 
+# Temporary fix for change in chromedriver 115 download location.
+# After `selenium-webdriver` 4.11 is released, we can use just that
+# instead of `webdrivers`.
+Webdrivers::Chromedriver.required_version = "114.0.5735.90"
```
dirschn, line 22: Are the double quotes required here? 😋

werebus:
> python versioning mismatch now? or something it looks like?

Interestingly, comparing that build to a previous one:

* Downloading the pre-built node failed for... some reason.
* nvm fell back to downloading source instead
* The `./configure` script for Node.js, which we don't usually run, needs Python and the default Travis VM doesn't actually have it.

I'm about to fix some single quotes (Calm down, codeclimate) so here's hoping it was a transient network problem.

benmelz (approved):

werebus:
After all that, looks like #290 works fine and seems like a better solution.

PR #290, opened 2023-07-20
Switch from webdrivers to just using selenium

werebus:
The internet tells me that new Selenium can manage your webdriver for you, too. This version doesn't know about the new download location for chromedriver 115 either, but it can download 114 and use that with Chrome 115.

This is an alternative to #289, and if Travis builds it successfully, I would say it's preferable.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -52,8 +52,8 @@ group :test do
   gem 'rack_session_access'
   gem 'rails-controller-testing'
   gem 'rspec-html-matchers'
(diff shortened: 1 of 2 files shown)


werebus:
Just one warning from Selenium about how it's using webdrivers 114 with Chrome 115. That warning will go away once selenium-webdriver 4.11 is release and we update.

benmelz (approved):

PR #302, opened 2023-08-24
AM/EVE split serverless functions

### LINKED ISSUE #285: Time dependent roster-picker
Transit operations is looking to do something more complicated with their on-call rotation and to have different rosters for "early" and "late".

We've had some discussions about a drastic restructure that would allow more flexibility on switchover behavior, but for now we're going to implement this outside of the app. It's fairly trivial to do handler logic _within_ Twillio. Some links:

* https://www.twilio.com/docs/serverless/functions-assets/quickstart/time-of-day-routing
* https://www.twilio.com/docs/voice/twiml/redirect
* https://www.twilio.com/docs/messaging/twiml/redirect

If the Twillio function returns some TwiML that just contains something like:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<Response>
    <Redirect>
        https://our-site.example.com/rosters/123/twilio/call.xml
    </Redirect>
</Response>
```

That should be all that's needed.

werebus:
Closes #285 

These functions are already deployed to Twilio, and tied to the Ops number. Should get some testing at some point when it isn't bothering Pires too much.

A lot of this PR is just boilerplate generated by Twilio's CLI.


benmelz (approved):

benmelz:
Approval stands

PR #303, opened 2023-08-29
Implement per-roster switchover times

werebus:
With "morning only" on-call, now, the late afternoon switchover time can't really be a one-size-fits-all situation without confusing users.


app/models/assignment.rb
```diff
@@ -32,17 +32,23 @@ def notify(receiver, of:, by:)
 
   class << self
     def between(start_date, end_date)
-      where('start_date <= ? AND end_date >= ?', end_date, start_date)
+      where arel_table[:start_date].lteq(end_date).and(arel_table[:end_date].gteq(start_date))
```
benmelz, line 35:
```suggestion
      where start_date: nil..end_date, end_date: start_date..nil
```
This isn't requested per-se but if we're going for "get rid of ugly interpolations", I'd offer this before getting into arel

app/views/rosters/new.haml
```diff
@@ -8,4 +8,7 @@
     = f.label :name
     = f.text_field :name, class: 'form-control'
   .form-group
-    = f.submit 'Save', class: 'btn btn-primary'
+    = f.label :swtchover_time
```
benmelz, line 11:
```suggestion
    = f.label :switchover_time
```

app/models/assignment.rb
werebus: I did this in order to be able to pass in other Arel nodes (e.g. by way of `.current`). With this change:

```
Roster.last.assignments.current
/home/matt/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/activerecord-7.0.6/lib/active_record/connection_adapters/abstract/quoting.rb:25:in `quote': can't quote Arel::Nodes::NamedFunction (TypeError)
```

benmelz (approved):

PR #307, opened 2023-09-05
Store phone number in roster

werebus:
Purely informational, but hought it might be useful for umts/round-three#5509


app/views/rosters/index.haml
```diff
@@ -6,6 +6,7 @@
       %th Name
       %th Fallback
       %th Switchover
+      %th Number
```
dirschn, line 9: Any reason for 'Number' rather than 'Phone'?

werebus: Hmmm... nope. As I was writing this feature, I kept mixing up in my head which one I wanted to use in general. But they _should_ be consistent.

dirschn (approved):

PR #308, opened 2023-09-06
Add Friendly ID gem for rosters

### LINKED ISSUE #41: Friendly ID
Hu-mans are too weak to consume numbers... just as they are too weak to consume gagh.

werebus:
Closes #41

### DIFF
--- a/app/models/roster.rb
+++ b/app/models/roster.rb
@@ -3,9 +3,12 @@
 require 'csv'
 
 class Roster < ApplicationRecord
-  has_paper_trail
-  has_many :assignments, dependent: :destroy
+  extend FriendlyId
 
+  has_paper_trail
+  friendly_id :name, use: :slugged
+
+  has_many :assignments, dependent: :destroy
   has_many :memberships, dependent: :destroy
   has_many :admin_memberships, -> { where(admin: true) },
            class_name: 'Membership', inverse_of: :roster
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2023_08_28_055929) do
+ActiveRecord::Schema[7.0].define(version: 2023_09_05_201033) do
   create_table "assignments", charset: "utf8mb4", collation: "utf8mb4_unicode_520_ci", force: :cascade do |t|
     t.integer "user_id"
     t.date "start_date"
@@ -35,7 +35,9 @@ ActiveRecord::Schema[7.0].define(version: 2023_08_28_055929) do
     t.datetime "updated_at", precision: nil, null: false
     t.integer "fallback_user_id"
     t.integer "switchover", default: 1020, null: false
+    t.string "slug"
     t.index ["name"], name: "index_rosters_on_name", unique: true
+    t.index ["slug"], name: "index_rosters_on_slug", unique: true
   end
 
   create_table "users", charset: "utf8mb4", collation: "utf8mb4_unicode_520_ci", force: :cascade do |t|
--- a/spec/requests/assignments_controller_spec.rb
+++ b/spec/requests/assignments_controller_spec.rb
@@ -94,7 +94,7 @@ RSpec.describe AssignmentsController do
         expect(Assignment.find_by(id: roster.id)).to be_nil
       end
 
-      it { is_expected.to redirect_to("/rosters/#{roster.id}/assignments") }
+      it { is_expected.to redirect_to("/rosters/#{roster.to_param}/assignments") }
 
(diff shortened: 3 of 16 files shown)


db/schema.rb
```diff
@@ -35,7 +35,10 @@
     t.datetime "updated_at", precision: nil, null: false
     t.integer "fallback_user_id"
     t.integer "switchover", default: 1020, null: false
+    t.string "phone"
```
dirschn, line 38: Seems like this might be a bit mixed with #307 ?

app/models/roster.rb
```diff
@@ -4,10 +4,12 @@
 
 class Roster < ApplicationRecord
   include PhoneHelper
+  extend FriendlyId
 
   has_paper_trail
-  has_many :assignments, dependent: :destroy
+  friendly_id :name, use: %i[slugged finders]
```
dirschn, line 10: Are you using `:finders` just so you don't have to use the `friendly` scope? It seems to me if we're setting it up for the first time, we should use their current defaults rather than an old version's defaults.

werebus: Yeah, the duck-type-aficionado in me kinda likes that both things count as roster "id"s and you don't have to change as much code. And, when I [first read the comments in the initializer](https://github.com/norman/friendly_id/blob/a72fac90282ad1e95611eb4bb381dd9d6702209b/lib/friendly_id/initializer.rb#L39-L45) (which I pruned down a bit), it sounded like it was along the lines of "it might be incompatible with your data, but it's more performant."

On second read, it sounds like the gem authors maybe wish they hadn't done it that way? Or at least there's disagreement among said developers. I think I've come around to adding explicit `.friendly`s, though.

spec/system/assignment/index_spec.rb
```diff
@@ -65,7 +65,7 @@ def bg_variable(var)
     context 'when clicking on an empty day' do
       let(:date) { Time.zone.today.change(day: 14) }
       let(:new_path) do
-        new_roster_assignment_path roster_id: roster.id,
+        new_roster_assignment_path roster_id: roster,
```
dirschn, line 68: Is this some feature of `friendly_id`? Shouldn't this be `roster: roster` or `roster_id: roster.to_param`?

spec/requests/assignments_controller_spec.rb
```diff
@@ -117,7 +117,7 @@
         expect(flash[:errors]).to eq('Only roster admins may delete assignments.')
       end
 
-      it { is_expected.to redirect_to("/rosters/#{roster.id}/assignments/#{assignment.id}/edit") }
+      it { is_expected.to redirect_to("/rosters/#{roster.to_param}/assignments/#{assignment.to_param}/edit") }
```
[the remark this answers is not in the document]
werebus: Yes, figured why not?

spec/system/assignment/index_spec.rb
werebus: `to_param` gets called on everything passed into route helpers. So, without FriendlyId, these are equivalent.

But the key needs to match the route fragment name, e.g.

```
       new_roster_assignment GET    /rosters/:roster_id/assignments/new(.:format)   assignments#new
```

But, the reason it looks unfamiliar is probably because, by virtue of being a resource route, this also works:

 ```suggestion
         new_roster_assignment_path roster,
```

To sum up:

```ruby
r = Roster.first
app.new_roster_assignment_path(roster: r)
# /home/matt/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/actionpack-7.0.7.2/lib/action_dispatch/journey/formatter.rb:44:in `path':
# No route matches {:action=>"new", :controller=>"assignments", :roster=>#<Roster id: 1, name: "Transit IT", created_at: "2023-09-01 07:33:34.000000000 -0400", updated_at: "2023-09-06 10:16:25.000000000 -0400", fallback_user_id: 4, switchover: 1000, phone: nil, slug: "transit-it">},
# missing required keys: [:roster_id] (ActionController::UrlGenerationError)

app.new_roster_assignment_path(roster_id: r)
#=> "/rosters/transit-it/assignments/new"
app.new_roster_assignment_path(roster_id: r.to_param)
#=> "/rosters/transit-it/assignments/new"
app.new_roster_assignment_path(roster_id: r.id)
#=> "/rosters/1/assignments/new"
app.new_roster_assignment_path(roster_id: r.id.to_param)
#=> "/rosters/1/assignments/new"
app.new_roster_assignment_path(r)
#=> "/rosters/transit-it/assignments/new"
app.new_roster_assignment_path(r.to_param)
#=> "/rosters/transit-it/assignments/new"
app.new_roster_assignment_path(r.id)
#=> "/rosters/1/assignments/new"
```

dirschn (approved):

PR #310, opened 2023-09-06
Use Rails encypted secrets

### LINKED ISSUE #264: Use `credentials` for secrets
In general this is something we should do, but _slightly_ more pressingly, the `secret_key_base` is set via an environment variable (stored somewhere on the server in the passenger config) which makes booting a production console difficult (you really just have to set the `SECRET_KEY_BASE="something"` environment variable, but that's not so obvious).

Should be rather similar to [jobapps#463](https://github.com/umts/jobapps/pull/463).

werebus:
`production.key` is on app3, and in the Password Database.

Closes #264


.gitattributes
```diff
@@ -0,0 +1,2 @@
+config/credentials/*.yml.enc diff=rails_credentials
+config/credentials.yml.enc diff=rails_credentials
```
dirschn, line 2: This is pretty cool

dirschn (approved):

PR #313, opened 2023-09-15
Don't allow admins to edit roster membership for other rosters

### LINKED ISSUE #292: Exception when editing a user to not be a member of the current roster anymore.
If you're editing a user from the "Roster A" route (e.g. if Roster A has an id of 1, '`/rosters/1/users/5`). And you don't include "Roster A" in the list of selected rosters, then you get an exception here:

https://github.com/umts/screaming-dinosaur/blob/334b3ba1630eb0fe283f3f26b66d94c9271b433a/app/controllers/users_controller.rb#L85-L86

Because, at that point, the user has already been updated to not be in that roster anymore, so `@user.membership_in @roster` is `nil`

Note that the intended action — removing the user from "Roster A" — is successful in this case. But the user gets a `500`.

### LINKED ISSUE #293: An admin from one roster can add and remove users from other rosters that they aren't admins in
An admin in one roster is just given a list of all the rosters to pick from when editing a user. Even if they aren't an admin in a particular roster, if they have access to edit users in **their** roster then they can add or remove that user's membership in that other roster.

dirschn:
closes #293 
closes #292 

I was going to clean up the create and edit actions both, but doing so would require some significant refactoring. I am still open to cleaning up both, but I wanted to make sure other people agreed it was necessary/ perhaps it could be a follow up so we can close this issue sooner.


app/controllers/users_controller.rb
```diff
@@ -8,7 +8,7 @@ class UsersController < ApplicationController
   WHITELISTED_ATTRIBUTES = [:first_name, :last_name, :spire, :email,
                             :phone, :active, :reminders_enabled,
                             :change_notifications_enabled,
-                            { rosters: [], membership: [:admin] }].freeze
+                            { roster_ids: [], membership: [:admin] }].freeze
```
dirschn, line 11: I did this because it was annoying me and I kept getting confused about whether I had rosters or ids. Turns out, most of that confusion stemmed from `parse_roster_ids`, but I think this change is warranted either way.

app/views/users/_form.html.haml
```diff
@@ -20,25 +20,36 @@
...
+          .list-group.list-group-form
+            - user_roster_options = @user.rosters | @current_user.memberships.where(admin: true).map(&:roster)
+            = f.collection_check_boxes :roster_ids, user_roster_options, :id, :name, include_hidden: false do |c|
+              - user_roster_permitted = @current_user.admin_in? c.object
+              = c.label class: ['m-0', 'list-group-item', user_roster_permitted ? 'list-group-item-action' : nil] do
+                .custom-checkbox.custom-control
+                  = c.check_box checked: @user.membership_in(c.object).present?,
+                                disabled: !user_roster_permitted,
+                                class: 'custom-control-input'
+                  = c.label class: 'custom-control-label'
```
werebus, line 35: HTML spec technically says that `<label>`s shouldn't have `<label>`s inside of them.

app/controllers/users_controller.rb
```diff
@@ -75,6 +72,19 @@ def destroy
...
 
+  def user_params
+    params.fetch(:user, {}).permit(
+      :first_name, :last_name, :spire, :email, :phone, :active, :reminders_enabled,
+      :change_notifications_enabled, roster_ids: [], membership: [:admin]
+    ).tap do |params|
+      given_roster_ids = params[:roster_ids].map(&:to_i)
+      params[:roster_ids] = (@user.rosters.map(&:id) || []).then do |roster_ids|
+        roster_ids.reject! { |roster_id| !roster_id.in?(given_roster_ids) && @current_user.admin_in?(roster_id) }
+        roster_ids | (given_roster_ids & @user.rosters.map(&:id).to_a)
```
werebus, line 83: This should be

```ruby
roster_ids | (given_roster_ids & whichever_rosters_the_current_user_is_admin_in)
```

Right? As is, I think this prevents adding a user to a roster.

werebus (changes requested):

app/controllers/users_controller.rb
```diff
@@ -75,6 +72,19 @@ def destroy
 
   private
 
+  def user_params
+    params.fetch(:user, {}).permit(
+      :first_name, :last_name, :spire, :email, :phone, :active, :reminders_enabled,
+      :change_notifications_enabled, roster_ids: [], membership: [:admin]
+    ).tap do |params|
+      given_roster_ids = params[:roster_ids].map(&:to_i)
+      params[:roster_ids] = (@user.rosters.map(&:id) || []).then do |roster_ids|
```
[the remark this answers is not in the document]
dirschn: I didn't think I could do this, since it's a `has_many` on this side?

app/controllers/users_controller.rb
```diff
@@ -83,21 +93,14 @@ def update_membership(membership_params)
     return true unless membership_params.present? && @current_user.admin_in?(@roster)
 
     membership = @user.membership_in @roster
+    return true if membership.nil?
```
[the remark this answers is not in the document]
dirschn: Yes, likely. That would be the reason I added this particular line.

app/views/users/_form.html.haml
dirschn: That may have to be a larger conversation, this was at @benmelz suggestion, following the use of this format multiple times elsewhere

app/controllers/users_controller.rb
dirschn: The reason I didn't worry about adding to a roster is because creating a user works fine, and there's a "add user from other rosters" input on the manage users page. When editing users, they don't show up as editable unless they're already in the roster you're an admin for.

app/views/users/_form.html.haml
```diff
@@ -20,25 +20,36 @@
           = f.phone_field :phone,
            placeholder: '+14135450056', class: 'form-control'
       .col-lg-6
+        = field_set_tag nil, class: 'form-group' do
+          %legend.m-0 Rosters
+          = f.hidden_field :'roster_ids[]', value: '', id: nil
+          .list-group.list-group-form
+            - user_roster_options = @user.rosters | @current_user.memberships.where(admin: true).map(&:roster)
```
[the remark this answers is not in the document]
dirschn: Okay, list all rosters and disable them unless you're an admin. Sure.

app/controllers/users_controller.rb
```diff
@@ -24,10 +24,9 @@ def new
   def edit; end
 
   def create
-    user_params = params.require(:user).permit(*WHITELISTED_ATTRIBUTES)
-    membership_params = user_params[:membership]
-    user_params = parse_roster_ids(user_params.except(:membership))
-    @user = User.new(user_params)
+    create_user_params = params.require(:user).permit(*WHITELISTED_ATTRIBUTES)
```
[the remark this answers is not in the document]
dirschn: Yeah. I started working on that but stopped because I was running into an issue when attempting to pull the new user's rosters in `user_params`, and in `update_membership`. I'll play with it though.

app/controllers/users_controller.rb
werebus: Ok, but I can still click the checkbox. It looked broken to me. In that, I checked off a roster and that did noting.

dirschn: What do you mean when you say you can click it? Meaning you can access the user's edit page when they're not in your roster?

app/controllers/users_controller.rb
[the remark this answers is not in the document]
dirschn: Yeah, I just tried in the console. I don't know where I got the idea that it doesn't work with a `has_many`.

app/controllers/users_controller.rb
werebus: No, I'm an admin in two rosters. I get to the user's edit page through roster 1, roster 2 is check-able because I'm an admin in roster 2. I check that box, save, no change to the user.

werebus:
Also, Travis failures seem legit.

app/views/users/_form.html.haml
werebus: Yeah, I mean, that's why I said "technically". But:

> **Permitted content**: Phrasing content, but no descendant `label` elements. No labelable elements other than the labeled control are allowed. 

Surrounding a checkbox with a label is fine, but you shouldn't need a second label inside. This seems to work:

```suggestion
                  .custom-control-label= c.object.name
```

werebus: Slightly better

```suggestion
                  .custom-control-label= c.text
```

app/views/users/_form.html.haml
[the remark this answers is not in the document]
dirschn: Do you have preferences about how to order? the obvious is by name, but would you like me to have rosters you're an admin in at the top and the others below? Or is it fine to have them mixed throughout?

dirschn:
> Also, Travis failures seem legit.

The specs are being weird, I'll see what I can do but might end up translating to request specs.

spec/controllers/users_controller_spec.rb
```diff
@@ -7,7 +7,14 @@
...
 
     let :submit do
-      post :create, params: { user: attributes, roster_id: roster.id }
+      post :create, params: {
+        user: attributes.merge({ roster_ids: [roster.id] }) do |k, o, n|
+          next unless k == :roster_ids
+
+          o.concat(n)
+        end,
+        roster_id: roster.id
```
dirschn, line 16: `attributes_for :user` contains `roster_ids` already, so I had to get a little fancy here

spec/factories/users.rb
```diff
@@ -7,6 +7,6 @@
     sequence(:spire) { |n| format('%08d@umass.edu', n) }
     sequence(:email) { |n| "user#{n}@umass.edu" }
     sequence(:phone) { |n| format('+1413545%04d', n) }
-    rosters { [create(:roster)] }
+    roster_ids { [create(:roster).id] }
```
dirschn, line 10: using roster_ids now, not rosters

werebus (changes requested):

werebus (approved):

werebus:
Man, my bad. That should have been a `<<` not a `+=`.

dirschn:
> Man, my bad. That should have been a << not a +=.

I suppose that's what I get for blindly trusting

PR #402, opened 2024-10-29
Fix paper trail whodunnit

werebus:
Accidentally broke setting of PaperTrail whodunnit in #368

I remember thinking, "Why is [`current_user` an `attr_accessor`](https://github.com/umts/screaming-dinosaur/pull/368/files#diff-766c34fd6533171eaf54300c153f89d6002c35c02cfc9c5b219251f85180ad07L4)? We never use it like that." Whoops.


spec/spec_helper.rb
```diff
@@ -10,6 +10,7 @@
 require 'rspec/rails'
 require 'rack_session_access/capybara'
 require 'umts_custom_matchers'
+require 'paper_trail/frameworks/rspec'
```
werebus, line 13: While we're here (I was making sure we actually have some tests for the "undo" feature).

benmelz (approved):

PR #433, opened 2024-12-20
Assignment Generator has Correct Default Start Date

alocke8181:
Closes #316 

The function `next_rotation_start_date` was moved to the `rosters` model and now only queries from the `assignments` of the current roster. Before it was querying from all rosters, regardless of whether it was for IT or OPS. `rotation_generator` in `assignments_controller` has been updated to reflect this.

Issue #423, opened 2024-12-11
Remove umts-custom-matchers


werebus:
This app is (will be shortly) the only one of ours in production that uses our old `umts-custom-matchers` gem (which only contains one matcher: `redirect_back`).

In order to upgrade to Rails 7.1, rspec-rails needs to be updated to 7.x, which is [prohibited by our gem](https://github.com/umts/custom-matchers/blob/3ebfc0f4c2f3ad344b8ca6a5447648344823c1fd/umts-custom-matchers.gemspec#L48).

PR #436, opened 2024-12-27
End_date Must be Before Start_date

### LINKED ISSUE #435: End Date can be Before Start Date
I was working on moving some of the spec's from `controllers/assignments_controller_spec` to `requests/assignments_spec` and for one of them, I tried setting the end date to be before the start date. When it's submitted it appears to be successful. I also tried doing it by running the development server and submitting redirects to the home page with the success flash message.

Looking through everything, it appears only the `weekday_generator` model checks that the end date is after the start date. Could someone else take a look to make sure I'm not crazy?

alocke8181:
Closes #435 

The end date for assignment generation must now be before the start date. I caught this bug while testing for PR #433. It was possible to generate an assignment with an end date before the start date. The assignments model now validates that the end date is greater than the start date, and `generate_rotation` in the controller also checks this. If the end date is before the start date, it redirects back and flashes an error message.


spec/controllers/assignments_controller_spec.rb
```diff
@@ -166,6 +170,19 @@
           .to redirect_to roster_assignments_path(roster, date: Time.zone.today)
       end
 
+      context 'when the end date is before the start date' do
+        before { params[:end_date] = Date.yesterday.to_fs(:db) }
```
benmelz, line 174: It's generally bad practice to "mutate" test setup data in `before` blocks - it's a pattern that can quickly get out of hand and produce some quirky behavior down the line.

Instead, what would be best for this example is for you to memoize `end_date` individually, much like `user_ids` and `starting_user_id` above.

spec/controllers/assignments_controller_spec.rb
```diff
@@ -166,6 +170,19 @@
...
+
+        it 'warns the end date is before the start date' do
+          submit
+          expect(flash[:errors]).not_to be_empty
+        end
+
```
benmelz, line 180:
```suggestion
```
We generally don't test flash messages at this level. Flash messages fall under "page content" which is more of a fit for system specs (you'll only see other examples here because this is an old controller spec).

benmelz (changes requested):
Expanding usages of `redirect_back` here, but I think that's fine as long as we get this in before we close out #423

benmelz (approved):

PR #441, opened 2025-01-02
Add HamlLint to Overcommit

### LINKED ISSUE #440: Add haml-lint

werebus:
Closes #440, I guess?

@benmelz, not sure what you ran into in #431, but `haml-lint` is [in the `Gemfile`](https://github.com/umts/screaming-dinosaur/blob/91918342d23c2a8544aea2b4a2bd190dbf9b724a/Gemfile#L36). This _seems_ to work OK for me?

### DIFF
--- a/.overcommit.yml
+++ b/.overcommit.yml
@@ -9,6 +9,11 @@ PreCommit:
   ForbiddenBranches:
     enabled: true
     branch_patterns: ["master", "main"]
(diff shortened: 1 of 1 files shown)


benmelz (approved):
I have no idea why I didn't think that haml-lint was available back then - so yeah we'll say this works

PR #442, opened 2025-01-03
Removed (Almost) All Uses of redirect_back from AssignmentsController

alocke8181:
Part of #423 

Removed nearly all occurrence of `redirect_back` and `report_errors` from `AssignmentsController`, and replaced them with either `redirect_to` or `render`. The only place it's still used is in `create`, since this can be called from both `rotation_generator` and `generate_by_weekday`. Do we want to have it simply redirect to the main page?


app/controllers/assignments_controller.rb
```diff
@@ -35,14 +35,14 @@ def generate_rotation
     user_ids = params.require :user_ids
     start_user = params.require :starting_user_id
     if end_date.before? start_date
-      flash[:errors] = 'The end date must be after the start date.'
-      redirect_back(fallback_location:
-                    roster_assignments_path(@roster)) and return
+      flash.now[:errors] = t('.end_before_start')
+      rotation_generator
```
benmelz, line 39: So if `rotation_generator` did a lot of setup for the form that would be hacky to replicate here, I would accept this (theoretically) lesser hack to get things working for now.

That being said, all that method does is set `@start_date` to prepopulate the field for the form. Conveniently we also accept start date as a param above.

Therefore I think it's less weird and even _slightly_ better to just change the `start_date` here to `@start_date` and reuse it for the form.

app/controllers/assignments_controller.rb
```diff
@@ -35,14 +35,14 @@ def generate_rotation
...
+      render :rotation_generator, status: :unprocessable_entity and return
     end
     unless user_ids.include? start_user
-      flash[:errors] = 'The starting user must be in the rotation.'
-      redirect_back(fallback_location:
-                    roster_assignments_path(@roster)) and return
+      flash.now[:errors] = t('.start_not_in')
+      rotation_generator
```
benmelz, line 44: Same as above

app/controllers/assignments_controller.rb
```diff
@@ -110,8 +111,8 @@ def destroy
       confirm_change(@assignment)
       redirect_to roster_assignments_path(@roster)
     else
-      flash[:errors] = t('.not_an_admin')
-      redirect_to edit_roster_assignment_path(@roster, @assignment)
+      flash.now[:errors] = t('.not_an_admin')
+      render :edit
```
benmelz, line 115:
```suggestion
      flash[:errors] = t('.not_an_admin')
      redirect_to edit_roster_assignment_path(@roster, @assignment)
```

This one was fine actually - `destroy` actions don't usually have a _dedicated_ form page or form fields so they'll _generally_ always redirect back to where they came from.

app/controllers/assignments_controller.rb
```diff
@@ -172,7 +173,7 @@ def require_taking_ownership
     return true if Current.user.admin_in?(@roster) || taking_ownership?
 
     flash[:errors] = t('.not_an_admin')
-    redirect_back fallback_location: roster_assignments_path(@roster)
+    redirect_to roster_assignments_path(@roster)
```
benmelz, line 176: Now this one needs a bit more thought because it's shared between `create` and `update` and ideally (based on how this is currently set up), we do still want to `render` over `redirect`ing, but with their respective form pages.

The easiest solution would be to parameterize the template we want to render here - i.e.

```ruby
def require_taking_ownership(error_template: nil)
  ...
  flash.now[:error] = t('.not_an_admin')
  render error_template, status: :unprocessable_entity
  ...
end
```
And then in the controller actions
```ruby
require_taking_ownership(error_template: :new) or return
```

spec/controllers/assignments_controller_spec.rb
```diff
@@ -74,12 +74,13 @@
...
-          expect { submit }.to redirect_back
+          submit
+          expect(flash[:errors]).not_to be_empty
         end
 
-        it 'redirects back' do
+        it 'redirects' do
           submit
-          expect(flash[:errors]).not_to be_empty
+          expect(response).to have_http_status(:found)
```
benmelz, line 83:
```ruby
          expect(response).to redirect_to(wherever_were_supposed_to_go_path)
```
Redirects have their own matcher that also checks the path that we're supposed to go

spec/controllers/assignments_controller_spec.rb
```diff
@@ -174,8 +175,9 @@
       context 'when the end date is before the start date' do
         let(:end_date) { Date.yesterday.to_fs(:db) }
 
-        it 'redirects back' do
-          expect { submit }.to redirect_back
+        it 'stays on the the generator page' do
+          submit
+          expect(response).to render_template :rotation_generator
```
benmelz, line 180:
```suggestion
          expect(response).to have_http_status(:unprocessable_entity)
```
Another case of "this is how things used to be done with controller actions but in modern day we keep things focused on the HTTP request/response itself"

spec/controllers/assignments_controller_spec.rb
```diff
@@ -187,8 +189,9 @@
           expect(flash[:errors]).not_to be_empty
         end
 
-        it 'redirects back' do
-          expect { submit }.to redirect_back
+        it 'stays on the generator page' do
+          submit
+          expect(response).to render_template :rotation_generator
```
benmelz, line 194:
```suggestion
          expect(response).to have_http_status(:unprocessable_entity)
```

spec/controllers/assignments_controller_spec.rb
```diff
@@ -424,8 +427,9 @@
           expect(flash[:errors]).not_to be_empty
         end
 
-        it 'redirects back' do
-          expect { submit }.to redirect_back
+        it 'stays on the edit page' do
+          submit
+          expect(response).to render_template :edit
```
benmelz, line 432:
```suggestion
          expect(response).to have_http_status(:unprocessable_entity)
```

spec/requests/assignments_controller_spec.rb
```diff
@@ -115,7 +115,10 @@
         expect(flash[:errors]).to eq('Only roster admins may delete assignments.')
       end
 
-      it { is_expected.to redirect_to("/rosters/#{roster.to_param}/assignments/#{assignment.to_param}/edit") }
+      it 'renders the edit page again' do
+        submit
+        expect(response).to render_template :edit
```
benmelz, line 120:
```suggestion
        expect(response).to have_http_status(:unprocessable_entity)
```

benmelz (changes requested):
> The only place it's still used is in create, since this can be called from both rotation_generator and generate_by_weekday. Do we want to have it simply redirect to the main page?

I'm not so sure this is true. AFAIK the only page that calls `create` is the `new` page (therefore it should rerender that form).

app/controllers/assignments_controller.rb
```diff
@@ -35,14 +35,14 @@ def generate_rotation
     user_ids = params.require :user_ids
     start_user = params.require :starting_user_id
     if end_date.before? start_date
-      flash[:errors] = 'The end date must be after the start date.'
-      redirect_back(fallback_location:
-                    roster_assignments_path(@roster)) and return
+      flash.now[:errors] = t('.end_before_start')
+      @start_date = start_date
```
benmelz, line 39:
```suggestion
```
If we just change `start_date` => `@start_date` on L33, L47 and L52, we don't have to add these at all

benmelz (changes requested):

app/controllers/assignments_controller.rb
```diff
@@ -73,29 +71,31 @@ def create
                        .permit :start_date, :end_date,
                                :user_id, :roster_id
     assignment = Assignment.new ass_params
```
benmelz, line 73: This needs to become `@assignment` for the form to rerender correctly on error, and this action needs to be added to the `set_roster_users` before action list

benmelz (changes requested):
Last thing

benmelz (approved):

PR #443, opened 2025-01-03
Removed All Occurrences of redirect_back and report_errors from RosterController

alocke8181:
Continuation of #423 

Removed all occurrences of `redirect_back` and `report_errors` from `RosterController` and the specs. `create` and `update` now re-render `:new` and `:edit` on failure with an `:uprocessable_entity` status. The specs that checked for a `redirect_back` now look for that status.

I noticed that `destroy` does not have any checks for errors besides `require_admin_in_roster`. All the other controllers have some form of a catch for errors when destroying. I'll create a small PR and add it after this one.

### DIFF
--- a/app/controllers/rosters_controller.rb
+++ b/app/controllers/rosters_controller.rb
@@ -4,7 +4,6 @@ class RostersController < ApplicationController
   api_accessible only: :show
 
   before_action :find_roster, only: %i[destroy edit setup show update]
-  before_action :find_users, only: %i[update]
   before_action :require_admin, except: %i[assignments show]
   before_action :require_admin_in_roster, only: %i[destroy edit setup update]
 
@@ -23,9 +22,7 @@ class RostersController < ApplicationController
     end
   end
 
-  def edit
-    @users = @roster.users
-  end
+  def edit; end
 
   def create
     @roster = Roster.new roster_params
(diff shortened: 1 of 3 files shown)


alocke8181:
I realize it might not be necessary to add the error handling in `destroy`. The `ApplicationController` will always set a roster before every action, so the only way the destroy would fail is if there are no rosters at all. Any other issues are caught by the `before_actions`.

app/controllers/rosters_controller.rb
```diff
@@ -44,7 +45,8 @@ def update
       confirm_change(@roster)
       redirect_to rosters_path
     else
-      report_errors(@roster, fallback_location: rosters_path)
+      flash.now[:errors] = @roster.errors.full_messages
+      render :edit, status: :unprocessable_entity
```
benmelz, line 49: This errors out due to a missing `@users` instance variable. Probably a `before_action` that isn't running.

benmelz (changes requested):

app/controllers/rosters_controller.rb
alocke8181: I've added `before_action :find_users` for `update`.

benmelz (approved):

PR #444, opened 2025-01-03
Removed All Occurrences of redirect_back from UsersController

alocke8181:
Continuation of #423 

Removed all occurrences of `redirect_back` and `report_errors` from `UsersController` and the specs. Actions now render the forms again and flash error messages.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -41,7 +41,7 @@ class UsersController < ApplicationController
       end
     else
       flash.now[:errors] = @user.errors.full_messages
-      render 'edit'
+      render :edit
     end
   end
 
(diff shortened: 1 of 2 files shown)


app/controllers/users_controller.rb
```diff
@@ -51,7 +51,8 @@ def transfer
       confirm_change(@user, "Added #{@user.full_name} to roster.")
       redirect_to roster_users_path(@roster)
     else
-      report_errors(@user, fallback_location: roster_users_path)
+      flash.now[:errors] = @user.errors.full_messages
+      render :index, status: :unprocessable_entity
```
benmelz, line 55: This one is weird since it's a single-field form on the index and doesn't have a dedicated "form" page. Meaning it's actually more appropriate to keep this as a redirect (but hardcoded to go back to the index)

benmelz (changes requested):

spec/controllers/users_controller_spec.rb
```diff
@@ -279,7 +279,8 @@
         end
 
         it 'redirects back' do
-          expect { submit }.to redirect_back
+          submit
+          expect(response).to have_http_status :unprocessable_entity
```
benmelz, line 283: This should now be failing but... isn't? But anyways should be testing with `redirect_to`

benmelz (changes requested):

PR #445, opened 2025-01-03
ChangesController Spec Does Not Use Redirect_Back

alocke8181:
More of #423 

The `ChangesController` spec no longer uses `redirect_back`. Instead, it just checks for the `:redirected` status.


benmelz (changes requested):
So for these ones we do want to completely reimplement the `redirect_back` matcher, meaning we want to set the `HTTP_REFERER` variable and make sure it redirects back to that path.

```ruby
before { request.env['HTTP_REFERER'] = '/redirect-target' }

it 'redirects back' do
  submit
  expect(response).to redirect_to('/redirect-target')
end
```

benmelz (approved):

PR #448, opened 2025-01-06
RotationGenerator Service Object

### LINKED ISSUE #439: Convert AssignmentsController#generate_rotation to service object
Like the weekday generator. Crafting this for @alocke8181 specifically (talk to me to go over service objects before you start working on this).

alocke8181:
Closes #439 

This creates a `RotationGenerator` service object to replace most of `generate_rotation` in the `AssignmentsController`. It now closely mirrors how assignments are generated by weekdays.


app/models/assignment/rotation_generator.rb
```diff
@@ -0,0 +1,60 @@
...
+    attribute :starting_user_id, :integer
+    attribute :start_date, :date
+    attribute :end_date, :date
+
+    validates :roster, presence: true
+    validates :user_ids, presence: true, length: { minimum: 1 }
+    validate :includes_start_user
+    validates :start_date, presence: true
+    validates :end_date, presence: true, comparison: { greater_than_or_equal_to: :start_date }
+
```
benmelz, line 19: `WeekdayGenerator` is a bad example for this, but good practice would be to make any method that isn't used externally `private` - I think that'd be everything here except for `def generate`

app/models/assignment/rotation_generator.rb
```diff
@@ -0,0 +1,60 @@
...
+    attribute :user_ids
+    attribute :starting_user_id, :integer
+    attribute :start_date, :date
+    attribute :end_date, :date
+
+    validates :roster, presence: true
+    validates :user_ids, presence: true, length: { minimum: 1 }
+    validate :includes_start_user
+    validates :start_date, presence: true
+    validates :end_date, presence: true, comparison: { greater_than_or_equal_to: :start_date }
```
benmelz, line 18:
```suggestion
    validates :roster, presence: true
    validates :user_ids, presence: true
    validates :starting_user_id, presence: true
    validates :start_date, presence: true
    validates :end_date, presence: true, comparison: { greater_than_or_equal_to: :start_date }
    validate :includes_start_user
```
Couple of things:
- `presence: true` should cover the `minimum: 1` condition since `[].present? => false`
- validate that `starting_user_id` is present separately
- keep custom validation declarations separate from built-in ones

app/models/assignment/rotation_generator.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    def roster
```
benmelz, line 20:
```suggestion
    def user_ids=(value)
      super(value&.map(&:to_i))
    end

    def roster
```

Here we can add a normalization step to the `user_id=` setter method. This means that the model will automatically cast the ids to integers and we won't have to be careful about type conversions (the unsightly `.to_s` calls) when working with this array

app/models/assignment/rotation_generator.rb
```diff
@@ -0,0 +1,60 @@
...
+      @roster ||= Roster.find_by(id: roster_id)
+    end
+
+    def includes_start_user
+      return unless user_ids
+
+      errors.add(:starting_user_id, 'is not included in the list') unless
+        user_ids.include? starting_user_id.to_s
```
benmelz, line 28:
```suggestion
      return if user_ids.blank? || starting_user_id.blank?
      return if user_ids.include? starting_user_id

      errors.add :starting_user_id, 'is not included in the list'
```
Generally if we're using a modifier (single-line) form of `if` or `unless`, the whole thing should fit on one line.

Then going off that, the more "standard" way of writing these custom validation methods is to filter out our positive conditions first with guard clauses, then add our error at the end.

app/models/assignment/rotation_generator.rb
```diff
@@ -0,0 +1,60 @@
...
+        end
+      end
+    rescue ActiveRecord::RecordInvalid => e
+      errors.merge! e.record.errors
+      raise e
+    end
+
+    def send_notification(assignment)
+      assignment.notify :owner, of: :new_assignment, by: Current.user
+    end
```
benmelz, line 58:
```suggestion
    def generate!
      validate!
      [].tap do |assignments|
        ActiveRecord::Base.transaction { create_assignments_and_write_to!(assignments) }
        assignments.each { |assignment| assignment.notify :owner, of: :new_assignment, by: Current.user }
      end
    rescue ActiveRecord::RecordInvalid => e
      errors.merge! e.record.errors
      raise e
    end

    def create_assignments_and_write_to!(output)
      rotation = user_ids.rotate user_ids.index(starting_user_id)
      (start_date..end_date).each_slice(7).with_index do |week, i|
        output << Assignment.create!(
          roster: @roster,
          start_date: week.first,
          end_date: week.last,
          user_id: rotation[i % rotation.size]
        )
      end
```
Here we have come code golfing/some refactoring to fix the rubocop offense we have here due to the complexity.

Non-trivially, the reason for the whole passing a mutable array to the creation method is that we want to wait until after the transaction succeeds to start notifying people about their assignments.

If we do it in the transaction itself, it's possible that people will get notified for assignments that didn't actually get created (assuming something went wrong on the tail end of the transaction).

spec/requests/rotation_generators_controller_spec.rb
```diff
@@ -0,0 +1,128 @@
+# frozen_string_literal: true
+
+RSpec.describe 'RotationGeneratorsController' do
```
benmelz, line 3: These all should be going in the existing `assignments_spec.rb` file

config/locales/en.yml
```diff
@@ -46,6 +46,10 @@ en:
           attributes:
             end_date:
               greater_than_or_equal_to: "must be on or after start date"
+        assignment/rotation_generator:
+          attributes:
+            end_date:
+              greater_than_or_equal_to: "must be on or after start date"
```
benmelz, line 52: We can make this a little more general/shareable - I've left a lot already so I'm not gonna get into how we should do this but just leaving a note for later.

spec/requests/rotation_generators_controller_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+        { assignment_rotation_generator: { start_date: start_date,
+                                           end_date: 1.week.from_now.end_of_week(:sunday),
+                                           starting_user_id: user1.id,
+                                           user_ids: [user1.id, admin.id] } }
+      end
+
+      it 'creates new assignments' do
+        expect { submit }.to change(Assignment, :count).by(2)
+      end
+
```
benmelz, line 44:
```suggestion

it 'creates new assignments with the correct attributes' do
  submit
  expect(Assignment.last(3).collect(&:attributes)).to contain_exactly(
    a_hash_including(...),
    a_hash_including(...),
    a_hash_including(...)
  )
end

```

So these specs are a good start, however we do want to cover as many edge cases as possible and also cover our implementation of the actual rotation generation process.

I think we should be generating at least 3 assignments against 2 users, and the starting user shouldn't be the first entry in `user_ids` (to test the `.rotate` part), and the end date should end earlier than sunday (so that we can test that the last week will get cut off accordingly).

Check in with me if you need a more in depth explanation on how to set up these cases

app/models/assignment/rotation_generator.rb
benmelz: Oh, also, this message should be localized as well

benmelz (changes requested):
`Roster#generate_assignments` I believe can be removed now

app/controllers/assignments_controller.rb
```diff
@@ -29,39 +28,18 @@ def new
...
-    if @generator.generate
+    if @weekday_generator.generate
       flash[:message] = t('.success')
-      redirect_to roster_assignments_path(@roster, date: @generator.start_date)
+      redirect_to roster_assignments_path(@roster, date: @weekday_generator.start_date)
     else
-      flash.now[:errors] = @generator.errors.full_messages.to_sentence
+      flash.now[:errors] = @weekday_generator.errors.full_messages.to_sentence
       render :generate_by_weekday, status: :unprocessable_entity
     end
```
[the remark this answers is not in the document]
alocke8181: changed back

app/controllers/assignments/rotation_generators_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+        render :prompt, status: :unprocessable_entity
+      end
+    end
+
+    private
+
+    def initialize_rotation_generator
+      @generator = Assignment::RotationGenerator.new(roster_id: @roster.id, **generate_rotation_params)
+      @generator.start_date ||= @roster.next_rotation_start_date
+      @generator.end_date ||= @generator.start_date + 3.months
```
benmelz, line 25:
```suggestion
      default_start = @roster.next_rotation_start_date
      @generator = Assignment::RotationGenerator.new(roster_id: @roster.id,
                                                     start_date: default_start,
                                                     end_date: default_start + 3.months,
                                                     user_ids: @roster.users.active.collect(&:id),
                                                     **generate_rotation_params)
```
I keep going back and forth on whether I want to request this or not - I think I'm going to leave it as a _light_ suggestion.

I find it a little weird to apply the user-given parameters and _then_ backfill default values.

It's _almost_ a problem in the fact that you might expect on first read that if a user explicitly submitted the form without dates, the blank values would get overridden with these defaults. The fact that we're only submitting through HTML and that blank values _should_ come in as `''` saves the day.

---

`user_ids` addition further explained below

app/views/assignments/rotation_generators/prompt.html.haml
```diff
@@ -0,0 +1,27 @@
...
+  = form_with model: @generator,
+              url: roster_assignments_generate_rotation_path(@roster),
+              local: true do |f|
+    .row
+      .col-lg
+        .mb-3
+          = f.label :start_date, nil, class: 'form-label'
+          = f.date_field :start_date, class: 'form-control'
+        .mb-3
+          = f.label :end_date, nil, class: 'form-label'
```
benmelz, line 13:
```suggestion
          = f.label :start_date, class: 'form-label'
          = f.date_field :start_date, class: 'form-control'
        .mb-3
          = f.label :end_date, class: 'form-label'
```
Not your doing, but these aren't necessary

app/views/assignments/rotation_generators/prompt.html.haml
```diff
@@ -0,0 +1,27 @@
...
+          = f.collection_select :starting_user_id, @roster.users.active.order(:last_name),
+                                :id, :last_name, {}, class: 'form-select'
+      .col-lg
+        .mb-3
+          = f.label :user_ids, 'Users', class: 'form-label'
+          = f.select :user_ids,
+                     options_from_collection_for_select(@roster.users.active.order(:last_name),
+                                                        :id, :last_name, @roster.users.pluck(:id)),
+                     { include_hidden: false },
+                     multiple: true, size: 10, class: 'form-select', selected: @roster.users.active
```
benmelz, line 26:
```suggestion
                     multiple: true, size: 10, class: 'form-select'
```
This was previously an ok way to set the default as the previous form didn't exactly handle errors very well, however...

Now this means that upon validation error, any changes made to this field will be wiped out. A better way to set this default would be in the controller with the others.

spec/requests/assignments/rotation_generator_spec.rb
```diff
@@ -0,0 +1,179 @@
+# frozen_string_literal: true
+
+RSpec.describe 'Assignments Generate Rotation' do
```
benmelz, line 3:
```suggestion
RSpec.describe 'Assignment Rotation Generators' do
```

benmelz (changes requested):

app/models/assignment/rotation_generator.rb
```diff
@@ -0,0 +1,70 @@
...
+    attribute :starting_user_id, :integer
+    attribute :start_date, :date
+    attribute :end_date, :date
+
+    validates :roster, presence: true
+    validates :user_ids, presence: true
+    validates :starting_user_id, presence: true
+    validates :start_date, presence: true
+    validates :end_date, presence: true, comparison: { greater_than_or_equal_to: :start_date,
+                                                       message: :end_date_must_be_after_start }
```
benmelz, line 19:
```suggestion
    validates :end_date, presence: true, comparison: { greater_than_or_equal_to: :start_date,
                                                       if: -> { start_date.present? && end_date.present? },
                                                       message: :end_date_must_be_after_start }
```
Totally missed this - we'll get redundant/weird messages if we don't ignore cases where either/both are missing

benmelz (changes requested):

app/controllers/assignments/rotation_generators_controller.rb
```diff
@@ -20,9 +20,12 @@ def perform
...
 
     def initialize_rotation_generator
-      @generator = Assignment::RotationGenerator.new(roster_id: @roster.id, **generate_rotation_params)
-      @generator.start_date ||= @roster.next_rotation_start_date
-      @generator.end_date ||= @generator.start_date + 3.months
+      default_start = @roster.next_rotation_start_date
+      @generator = Assignment::RotationGenerator.new(roster_id: @roster.id,
+                                                     start_date: default_start,
+                                                     end_date: default_start + 3.months,
+                                                     user_ids: @roster.users.active.order(:last_name).pluck(:id),
```
benmelz, line 27:
```suggestion
                                                     user_ids: @roster.users.active.pluck(:id),
```
Ordering shouldn't matter here

benmelz (changes requested):

benmelz (approved):

PR #534, opened 2025-04-15
Rubocop models

werebus:
Thought I'd do a little linting, mostly a refactor of the overlapping validation.


app/models/roster.rb
```diff
@@ -10,10 +10,14 @@ class Roster < ApplicationRecord
...
   has_many :assignments, dependent: :destroy
   has_many :memberships, dependent: :destroy
+
+  # Covered by the dependent: :destroy on memberships
+  # rubocop:disable Rails/HasManyOrHasOneDependent
   has_many :admin_memberships, -> { where(admin: true) },
            class_name: 'Membership', inverse_of: :roster
   has_many :non_admin_memberships, -> { where.not(admin: true) },
            class_name: 'Membership', inverse_of: :roster
+  # rubocop:enable Rails/HasManyOrHasOneDependent
```
benmelz, line 20: Rubocop will also accept a `dependent: nil` option here as a confirmation of "yes I want to do nothing" for this cop (it's what we do on the hub for the flex models).

Maybe just a little more sightly than disabling the linter altogether?

```suggestion
  has_many :admin_memberships, -> { where(admin: true) },
           class_name: 'Membership', dependent: nil, inverse_of: :roster
  has_many :non_admin_memberships, -> { where.not(admin: true) },
           class_name: 'Membership', dependent: nil, inverse_of: :roster
```

app/models/user.rb
```diff
@@ -15,8 +15,7 @@ class User < ApplicationRecord
   validates :first_name, :last_name, :spire, :email, :phone, :rosters, presence: true
   validates :spire, :email, :phone, uniqueness: { case_sensitive: false }
   validates :calendar_access_token, uniqueness: { case_sensitive: true }
-  validates :spire, format: { with: /\A\d{8}@umass.edu\z/,
-                              message: 'must be 8 digits followed by @umass.edu' }
+  validates :spire, format: { with: /\A\d{8}@umass.edu\z/ }
```
benmelz, line 18: Also another preference I've built up from doing this sort of thing a lot on the hub - it might be a little nicer in terms of future maintenance to do something like

```suggestion
  validates :spire, format: { with: /\A\d{8}@umass.edu\z/, message: :must_be_eight_digits_and_umass }
```

combined with

```yaml
en:
  ...
  errors:
    messages:
      must_be_eight_digits_and_umass: 'must be 8 digits followed by @umass.edu'
```

i.e. the explicit key here makes it easier to infer that this is overridden with a custom translation, as well as makes it much easier to find said translation (especially when the translation file gets bigger)

benmelz (approved):
Suggestions than requests

Issue #299, opened 2023-08-23
Admin CSV


sherson:
We discussed possibly having a dashboard for admins, though a CSV summary for a given roster might be more useful (can be easily Excelified for various purposes). These fields come to mind:

```
rosters.name
users.first_name
users.last_name
users.id (edit: replace this with users.email)
assignments.start_date
assignments.end_date
assignments.created_at
assignments.updated_at
```

benmelz:
> rosters.name

I'm thinking that it makes the most sense to offer this CSV in the scope of the roster (due to admin status being on a per-roster basis). As such this column is just gonna be uniformly the same value - might make more sense to just put the roster name in the file name?

> users.id

Would this have a use case? (in the vein of don't hand out database ids for stuff)

sherson:
So initially I drafted this issue as exactly that: just put the roster name in the file name.

However, the admin for the operations rosters may want to combine the AM and EVE rosters for whatever Excelery needs to be done, so it would probably be useful as a field.

sherson:
Just realized I forgot to answer your second question: `users.id` might be convenient for grouping/summing per user (there are ways to do it with two columns, but having a single unique column for each user is simpler).

sherson:
Considered suggesting spire, but I try not to use that unless necessary (it's sort-of-semi-sensitive). Could go with email instead if you'd rather avoid `users.id`.

sherson:
Now that I think about it, email might actually be useful for other things too, so let's use that instead.

PR #602, opened 2025-08-18
Fix CSV export when using friendly-id slugged ids

werebus:
@sherson asks for an "Admin CSV" in August of 2023 (#299), gets it. Then Matt breaks it 1 week later (#308) - remains broken for two years.


frothedoatmilk (approved):
Starting the clock now: we have one week to break this again

PR #629, opened 2025-09-17
Remove Timecop gem

werebus:
Rails has built-in time travel, generally no need for a separate gem.


config/environments/test.rb
```diff
@@ -54,15 +54,4 @@
   # RackSessionAccess gives us access to the rack session
   # during feature tests
   config.middleware.use RackSessionAccess::Middleware
-
```
werebus, line 57: This doesn't work well enough to keep: a bunch of system tests fail when Rails and Chrome disagree about what time it is. One can still do this when troubleshooting a particular test or whatever, but it's misleading for me to leave this here implying it'll just work.

spec/models/assignment_spec.rb
```diff
@@ -218,17 +218,17 @@
     end
 
     context 'when it is before the switchover' do
-      around do |example|
```
werebus, line 221: rubocop-rspec-rails complains about `travel_to` in an `around` hook. This is because the `TimeHelper` methods always return to the present at the end of the test (via an `after_teardown` hook).

benmelz (approved):

PR #643, opened 2025-09-18
Move to rollup via jsbundling

benmelz:
Look at those modules. Way trickier than cssbundling, but this gets us much closer to a modern js setup. I tried rollup this time instead of esbuild just due to maturity - in the end they end up being more or less equivalent in terms of complexity.


Gemfile.lock
frothedoatmilk, line 1: Can you run `bundle i` to generate the sha256 here?

frothedoatmilk (changes requested):

PR #728, opened 2025-12-12
Last two possibly-dangerous Rails 8.1 defaults

werebus:


config/initializers/new_framework_defaults_8_1.rb
```diff
@@ -58,7 +58,7 @@
 # Applications that want to allow these redirects can set the config to `:log` (previous default)
 # to only log warnings, or `:notify` to send ActiveSupport notifications.
 #++
-# Rails.configuration.action_controller.action_on_path_relative_redirect = :raise
+Rails.configuration.action_controller.action_on_path_relative_redirect = :raise
```
werebus: Double-checking this is what lead me to #727

I think it's possible that someone could hit this exception there doing something weird. But... that's probably what we want.

PR #775, opened 2026-01-13
Implement versions policy

LateNightFlowerField:


app/controllers/changes_controller.rb
```diff
@@ -2,9 +2,11 @@
 
 class ChangesController < ApplicationController
   before_action :find_version
-  before_action :require_original_user
+
+  def implicit_authorization_target = self.class.controller_path.to_sym
 
   def undo
+    authorize! context: { user_id: @version.whodunnit.to_i }
```
benmelz:
```suggestion
    authorize! @version
```

Controller isn't named as such, but for intents and purposes this is really a `VersionController`. As such I think it makes much more sense to just authorize against it as the `record`.

benmelz (changes requested):

benmelz (approved):

PR #778, opened 2026-01-13
Implement assignment policy

benmelz:


app/controllers/assignments_controller.rb
```diff
@@ -89,12 +77,17 @@ def assignment_params
...
     @assignment = Assignment.includes(:user).find(params.require(:id))
+    @previous_owner = @assignment.user
   end
 
   def set_roster_users
     @users = @roster.users.active.order :last_name
   end
 
+  def allow_calendar_token_access
+    Current.user ||= User.find_by(calendar_access_token: params[:token])
```
LateNightFlowerField: Something about setting `Current.user` outside of auth feels off to me, maybe instead use
```ruby
authorize! context: { user: User.find_by(calendar_access_token: params[:token])}
```
to override user for feed authorization.

app/controllers/assignments_controller.rb
```diff
@@ -108,6 +101,11 @@ def index_json
     render layout: false
   end
 
+  def index_csv
+    @roster = Roster.preload(assignments: :user).friendly.find(params[:roster_id])
+    render csv: @roster.assignment_csv, filename: @roster.name
```
LateNightFlowerField: Not sure what conventions are for transit repos, but a more descriptive filename could be useful (eg. "Transit IT Assignments Thu 1/18")

PR #780, opened 2026-01-13
Update tests for access redux changes

LateNightFlowerField:
Note:
Reverts change in (https://github.com/umts/screaming-dinosaur/pull/770/changes#diff-766c34fd6533171eaf54300c153f89d6002c35c02cfc9c5b219251f85180ad07L41) that restricted sessions to active users.

### DIFF
--- a/spec/system/users/index_spec.rb
+++ b/spec/system/users/index_spec.rb
@@ -1,32 +1,33 @@
 # frozen_string_literal: true
 
 RSpec.describe 'user index' do
-  let(:roster) { create :roster }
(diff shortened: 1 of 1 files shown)


LateNightFlowerField:
One of the tests relies on admin deactivating themselves, happy to update that test, but also run into the issue where the last admin is able to deactivate themself even though they cannot delete themself. More logic for deactivation would probably be needed.

LateNightFlowerField:
I'll update the test to use admin deactivating another user, and will look into preventing user self deactivation.

Issue #781, opened 2026-01-13
Add a real dashboard page


benmelz:
With the whole multiple rosters thing happening now, having a specific roster's assignments as the root page is a little clunky, so I think it's time to introduce some form of a dedicated homepage.

To start small, I'm thinking this should be our list of specs:
- if the user is only a part of one roster, just send them there still
- if the user is currently on call, note that and note which roster(s)
- list and link to the rosters a user is a part of

We can flesh it out as we go, but first PR should probably just be a bare minimum of these things

PR #782, opened 2026-01-14
Add a simple dashboard

LateNightFlowerField:
Part of #781
Redirect to assignment page if user is only in one roster.


LateNightFlowerField:
Forgot to track controller in git will remake it now.

benmelz (approved):
Tests I think are due to the semantic changes of root path, but they probably will be revisited as a part of the greater project, so I'm leaving them

PR #792, opened 2026-01-15
Fix/prevent admin self deactivate

KernSharma:
Prevents admins from deactivating their own accounts.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -258,7 +258,7 @@ GEM
     pp (0.6.3)
       prettyprint
     prettyprint (0.2.0)
-    prism (1.8.0)
+    prism (1.9.0)
     propshaft (1.3.1)
       actionpack (>= 7.0.0)
       activesupport (>= 7.0.0)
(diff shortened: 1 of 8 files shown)


config/database.yml.sample
```diff
@@ -2,7 +2,7 @@ default: &default
   adapter: trilogy
   encoding: utf8mb4
   collation: utf8mb4_unicode_520_ci
-  username: <%= ENV['USER'] %>
+  username: root
```
benmelz:
```suggestion
  username: <%= ENV['USER'] %>
```
While we probably should update this at some point, it doesn't really belong in _this_ PR

spec/models/user_spec.rb
```diff
@@ -52,4 +52,23 @@
...
+      user.active = false
+      expect(user).not_to be_valid
+    end
+
+    it 'adds error message when user tries to deactivate themselves' do
+      user.active = false
+      user.valid?
+      expect(user.errors[:base]).to include('You may not deactivate yourself')
+    end
+  end
```
benmelz:
```suggestion
  describe '#valid?' do
    subject(:call) { user.valid? }

    context 'when logged in as the subject and attempting deactivation' do
      around { |example| Current.set(user:) { example.run } }

      let(:user) { create :user }

      before { user.active = false }

      it 'does not allow you to deactivate yourself' do
        expect(call).to be(false)
      end

      it 'adds a helpful error message' do
        call
        expect(user.errors[:base]).to include('You may not deactivate yourself')
      end
    end
  end
```

This is more styling than anything, but this is more along the lines of how we typically structure/organize for rspec.

PR #794, opened 2026-01-16
New rosters index/dashboard page

LateNightFlowerField:


app/views/rosters/index.html.haml
```diff
@@ -1,4 +1,4 @@
-= render layout: 'layouts/section', locals: { heading: 'On-Call Dashboard' } do
+= render layout: 'layouts/section', locals: { heading: 'On-Call Dashboard', creatable: new_roster_path } do
```
benmelz:
```suggestion
= render layout: 'layouts/section',
         locals: { heading: 'On-Call Dashboard', creatable: allowed_to?(:new, Roster) && new_roster_path } do
```
As a part of this project we should probably be adding access guards to any/all links we can

PR #807, opened 2026-01-21
Destroy future assignments for non-members

### LINKED ISSUE #804: Delete future assignments when roster membership is deleted
Much like how they are deleted when users are deactivated, they should probably also go when memberships are destroyed.

zafnunhasnat-ops:
Closes #804 
Added a destroy method to delete future assignments for non-members.


app/models/membership.rb
```diff
@@ -14,4 +15,11 @@ def at_least_one_admin
 
     errors.add :user, 'is the last admin and cannot be demoted'
   end
+
+  def delete_future_assignment
```
benmelz:
```suggestion
  def delete_future_assignments
```

app/models/membership.rb
```diff
@@ -14,4 +15,11 @@ def at_least_one_admin
...
+    user.assignments
+        .where(roster: roster)
+        .upcoming
+        .destroy_all
```
benmelz:
```suggestion
    user.assignments.where(roster:).future.destroy_all
```

PR #822, opened 2026-01-23
Reorganize feed system 

### LINKED ISSUE #813: Reorganize feed system components
I'm kind of bikeshedding here but I think this would be a nice step-up issue for @zafnunhasnat-ops:

The oddness of the url construction for feeds, `twilio` already having its own controller as well as the fact that there actually is a backing model for "feeds" makes me want to move this endpoint out of `AssignmentsController`.

Much of the content for this will be refactoring, namely
- `/feed/:roster/:token` should remain the public path, but it should route to a new `FeedController#show` endpoint
- Move `lib/assignments_ics.rb` to the models directory - I think conveniently we can name it `feed.rb`

As well as any fallout from those changes (pay attention to policies and tests).

zafnunhasnat-ops:
Closes #813 
Refactor feed controller and handle policies and test cases.


app/controllers/assignments_controller.rb
```diff
@@ -107,7 +90,7 @@ def index_csv
   end
 
   def render_ics_feed
-    ics = AssignmentsIcs.new(@roster.assignments)
-    render plain: ics.output, content_type: 'text/calendar'
+    feed = Feed.new(@roster.assignments)
+    render plain: feed.output, content_type: 'text/calendar'
```
benmelz: This is no longer used since we moved the parent action to the `FeedController` - it can be removed now as well

app/controllers/feed_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+  def allow_calendar_token_access
+    return if params[:token].blank?
+
+    Current.user ||= User.find_by(calendar_access_token: params[:token])
+  end
+
+  def render_ics_feed
+    feed = Feed.new(@roster.assignments)
+    render plain: feed.output, content_type: 'text/calendar'
+  end
```
benmelz: Some of this is a bit redundant now - as a hint try to get rid of this method and implement everything as a part of the `show` action and you should see what I mean

benmelz (changes requested):

PR #829, opened 2026-01-29
Rework flash system

benmelz:
A couple of overall changes:
- constrains our flash variants to the built-in `notice` and `alert`
- backs message building with locales (through actions and model names)
- makes undoability opt-in

### DIFF
--- a/spec/system/users/index_spec.rb
+++ b/spec/system/users/index_spec.rb
@@ -79,7 +79,7 @@ RSpec.describe 'user index' do
 
       it 'informs you of success' do
         click_button 'Activate'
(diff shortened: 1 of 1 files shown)


spec/system/users/index_spec.rb
```diff
@@ -79,7 +79,7 @@
 
       it 'informs you of success' do
         click_button 'Activate'
-        expect(page).to have_css 'div', text: 'User has been updated.'
+        expect(page).to have_text 'Successfully updated user'
```
alocke8181:
```suggestion
        expect(page).to have_text('Successfully updated user')
```

Issue #872, opened 2026-02-20
Scalable user selects


benmelz:
Also far from needed currently, but since we will be removing the concept of inactivity from users, theoretically at least our "add membership" select will not scale (and be annoying to use since native select is not searchable).

@frothedoatmilk and I have been eyeing "tom-select" for round-three, perhaps this would be a good place to pilot it (supports searching and async options).

PR #879, opened 2026-02-25
Set up tom-select

benmelz:
Opens up #872

Sets up the lib with basic configuration, applies to the dev login page selects

### DIFF
--- a/app/javascript/controllers/tom_select_controller.js
+++ b/app/javascript/controllers/tom_select_controller.js
@@ -6,6 +6,7 @@ export default class extends Controller {
     new TomSelect(this.element, {
       create: false,
       plugins: ['dropdown_input'],
+      refreshThrottle: 0,
     });
   }
 }


app/javascript/controllers/tom_select_controller.js
```diff
@@ -0,0 +1,11 @@
+import {Controller} from '@hotwired/stimulus';
+import TomSelect from 'tom-select';
+
+export default class extends Controller {
+  connect() {
+    new TomSelect(this.element, {
+      create: false,
+      plugins: ['dropdown_input'],
```
frothedoatmilk:
```suggestion
      plugins: ['dropdown_input'],
      refreshThrottle: 0,
```
I think we can safely lower this, if we find it needs raising later we can implement a default, and this is closer to selectpicker live search behavior

frothedoatmilk (approved):

PR #950, opened 2026-03-30
Rework assignment model to store times

benmelz:
Closes #949

- [x] add validation that assignment end datetime can't be before/on roster created at
- [x] include roster created at as initial lag value in `Assignment.with_start_datetimes`
- [ ] handle am/eve merge in migration (?)


frothedoatmilk:
The new assignments index is a bit clunky, there's no header or other indication of what each column is:

image

The mobile view is fairly unfriendly as well:

image

We've dealt with similar mobile view issues before, specifically with list groups, but I can't remember which project that was on. Mobile view should probably confine the buttons to the bottom right of each element, and give space for the times.

PR #964, opened 2026-04-08
Don't offer undo when a record has no saved changes

### LINKED ISSUE #869: Undo may attempt unrelated previous version if no changes are saved
If you submit one of the REST forms with no changes, rails won't actually "save" anything and thus not create a version for it.

Furthermore if the corresponding controller marks the action as undoable, the current system will still offer an undo button for `versions.last`, which may be a completely unrelated previous version.

So we should probably make sure that we actually did something to the record before offering the undo button.

KernSharma:
Closes #869

Fixes a bug where submitting a form with no changes would still show an undo button, potentially pointing at an unrelated previous version.

### DIFF
--- a/coverage/.last_run.json
+++ b/coverage/.last_run.json
@@ -1,6 +1,6 @@
 {
   "result": {
-    "line": 40.93,
-    "branch": 47.05
+    "line": 35.75,
+    "branch": 25.96
   }
 }
--- a/coverage/.last_run.json
+++ b/coverage/.last_run.json
@@ -1,6 +1,6 @@
 {
   "result": {
-    "line": 35.75,
-    "branch": 25.96
+    "line": 97.66,
+    "branch": 85.29
   }
 }
--- a/coverage/.last_run.json
+++ b/coverage/.last_run.json
@@ -1,6 +1,6 @@
 {
   "result": {
-    "line": 97.37,
-    "branch": 84.55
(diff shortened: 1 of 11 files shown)


spec/requests/rosters_spec.rb
```diff
@@ -299,6 +299,29 @@
         expect { submit }.not_to(change { roster.reload.attributes })
       end
     end
+
+    context 'with versioning', :versioning do
```
benmelz: We generally don't like to test the flash hash directly in controlle/request specs, as it's beyond the scope of our "agent".

We should really convert this to a system spec instead.

coverage/.last_run.json
```diff
@@ -1,6 +1,6 @@
 {
   "result": {
-    "line": 97.37,
-    "branch": 84.55
+    "line": 40.93,
+    "branch": 47.05
```
benmelz: These are most likely erroneous - make sure that _if_ you want to commit the coverage file here, it's after a full run of the test suite.

package-lock.json
```diff
@@ -835,9 +835,9 @@
       }
```
benmelz, line 835: This file shouldn't be changed as a part of this PR

benmelz (changes requested):

benmelz (approved):

PR #979, opened 2026-04-18
Fix test suite to follow assignment DB(end_datetime)

zafnunhasnat-ops:
Fix specs to follow end_datetime correctly.


benmelz:
This was too much in one PR - leaving branch to copy contents over to others but in current form this won't be merged

PR #985, opened 2026-04-22
Downgrade and pin bigdecimal

benmelz:
Bundler deploy problem. I'm actually just guessing with this, but it failed to compile so it's probably this?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,8 +5,8 @@ ruby file: '.ruby-version'
 
 gem 'action_policy'
 gem 'bootsnap', require: false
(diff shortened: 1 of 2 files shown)


werebus:
I doubt this will fix your problem. `force_ruby_platform` forces compiling on the host _instead_ of downloading the platform-specific precompiled gem. I don't see that `bigdecimal` has any precompiled versions (except for jruby).

benmelz:
Yeah upon actual investigation it looks like 4.1.2 introduced syntax that requires `gcc >= 5` (we have 4)

werebus (approved):

PR #1010, opened 2026-05-06
Remove week and weekday assigner 

zafnunhasnat-ops:
Remove anything and everything related to week and weekday assigner.


benmelz (approved):
Awesome

PR #1023, opened 2026-05-18
Assignments index

zafnunhasnat-ops:
Closes #1009 
Adds index for assignments (start, end, user, edit, delete) separate from the calendar view.


app/views/layouts/_roster.html.haml
```diff
@@ -6,6 +6,10 @@
     %li.nav-item
       = link_to roster_path(roster), class: ['nav-link', nav_item == 'assignments' && 'active'] do
```
benmelz:
```suggestion
      = link_to roster_path(roster), class: ['nav-link', nav_item == 'overview' && 'active'] do
```
And then this has to be adjusted in `app/views/rosters/show.html.haml` as well

benmelz (commented):
Looks great so far, we should be adding some request specs as well

PR #1025, opened 2026-05-18
anchor assignment start datetimes

KernSharma:
Closes the two model-only items on the #950 checklist:                                                                                                                                                                                            - Adds a validation that an assignment's `end_datetime` must be after its roster's `created_at`. Without this, the LAG anchor below would be a meaningless point in time.  
- Passes `rosters.created_at` as the third argument to the `LAG()` window function in `Assignment.with_start_datetimes`, so the first assignment per roster resolves to a real `start_datetime` instead of `NULL`. The subquery now joins `rosters` so the column is in scope.

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -8,7 +8,9 @@ class Assignment < ApplicationRecord
   belongs_to :roster
   belongs_to :user, optional: true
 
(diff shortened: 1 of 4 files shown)


benmelz:
So both of these items are already completed on `benmelz/assignments` and were checked off two weeks ago...

Per our teams conversations your assignment was to work on reimplementing model callback logic (one at a time)

PR #1037, opened 2026-05-20
Refactor assignment notification callbacks specs

zafnunhasnat-ops:
Relates to #1033
Update assignment model spec to follow the new notification callbacks methods (notify_user_of_assignment, notify_user_of_change, notify_user_of_removal).


benmelz:
This one's already open on #1033

PR #1038, opened 2026-05-20
weekly reminders

KernSharma:
Replaces old daily 9am per-assignment reminder with a weekly sunday night digest. Instead of 1 email per assignment, each user gets a digest at 8pm EST Sunday listing all their upcoming assignments for the next week grouped by roster


app/models/assignment.rb
```diff
@@ -30,6 +30,12 @@ def with_start_datetimes
                      .as(arel_table.name)
     end
 
+    def send_reminders!
+      assignments_in_next_week.group_by(&:user).each do |user, assignments|
+        AssignmentsMailer.upcoming_reminder(user, assignments).deliver_now
+      end
+    end
```
benmelz: I can't see this logic being used anywhere except this specific job, so I think we should take the opportunity to move it there (same with the private helper)

benmelz (changes requested):

PR #1051, opened 2026-06-02
Render ICS feed events as timed datetimes

KernSharma:
Switches the ICS feed from all-day events to timed UTC datetimes using the new start_datetime/end_datetime fields.


benmelz:
This was done by @zafnunhasnat-ops in https://github.com/umts/screaming-dinosaur/pull/1046

PR #1057, opened 2026-06-04
Switch assignment mailer to datetime

zafnunhasnat-ops:
-Updated mailer methods/specs/related views to use start_datetime and end_datetime


benmelz:
@zafnunhasnat-ops Some gnarly merge conflicts here - maybe do #1058 first then tackle it all at once

PR #1062, opened 2026-06-08
Display and persist UPN from AD login

### LINKED ISSUE #968: Add some login clarifications
We had some confusion today from someone who had:
- registered with their main account
- at registration, changed their email to their `-af` email
- attempted to log back in with `-af` account and was bewildered why they were prompted to create an account (and could not due to email uniqueness constraint)

Face value I think we should
- add some clarifying text under the email field that it's strictly for contact purposes and in no way is linked to login
- display which account is currently being used on registration as a disabled field (should be available at that point in the session)

I'm not sure if we should try to record the email we get from AD on signup. We don't really have a use for it in normal operations but it would have come in handy for debugging this today?

KernSharma:
Closes #968 

Adds clarifying text under the email field so users don't mistake it for their login email, and shows the Microsoft account they're signed in with as a disabled field on registration. Persists the AD-supplied UPN on the User record by reading the 'upn' claim from omniauth-entra-id's raw_info on every login, refreshing it whenever it changes or is missing.


benmelz (changes requested):
One more thing I'd like to add while we're here - when editing a user...
- if the current user is an admin, display the entra_uid
- if the current user is an admin and not editing themselves, allow them to edit the entra_uid

spec/requests/users_spec.rb
```diff
@@ -69,6 +69,27 @@
         call
         expect(response).to be_successful
       end
+
+      context 'with a UPN in the session' do
```
benmelz: This example and the below ones that reference `Capybara` do not belong in request specs. We should be asserting against HTTP statuses here, not body content

benmelz (changes requested):

benmelz (approved):

PR #1079, opened 2026-06-16
Add assignment groups and taking

KernSharma:
Adds an AssignmentGroup model and an optional assignment_group association on Assignment, plus an AssignmentTaker service and prompt page that lets any logged-in user take an assignment (or the whole group at once when it's grouped), with a Take link on each row. It also fixes the uncovered_periods_between spec, which had gone time-brittle and started failing this week.

Closes #1067.


app/policies/assignment_taker_policy.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class AssignmentTakerPolicy < ApplicationPolicy
+  def manage? = user.present?
```
benmelz: This should be slightly narrowed to check if the user is a member of the roster (and ideally depend on an existing `RosterPolicy` or `AssignmentPolicy` implementation)

benmelz (changes requested):
- It might be helpful to also add a `:name` column (required) to help identify what the grouping represents
- When clicking on an assignment via the calendar it should probably send you to the new take page now

PR #1080, opened 2026-06-18
Add Assignment Generator

zafnunhasnat-ops:
Closes #1060 
Replaces week-day assigner with assignment generator.
Includes assignment generator routes, model, controller, view, policy, request spec,


benmelz (changes requested):
Looks great so far. The biggest thing that is missing right now is some logic tests for the actual generator behavior.

What I mean by that is that we should have at least one test case that has as many edge cases as possible included and we should be testing that the exact assignments that we expect get generated.

It would be appropriate to do this as a part of the request spec, or you could choose to implement it as a model spec, whatever you prefer.

benmelz (approved):

PR #1107, opened 2026-07-07
Rewrite seeds for datetime assignments

KernSharma:
Fixes the seeds assignments loop for the datetime refactor — assignments no longer have start_date/end_date, so `rails db:seed` crashed there. Each user now gets an end_datetime of `(i + 1).weeks.since.beginning_of_week(:friday)`, which reproduces the old Friday-to-Friday weekly rotation since start times derive from the previous assignment's end (or the roster's created_at for the first).

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -4,7 +4,8 @@ exit unless Rails.env.development?
 
 # ROTATIONS
 transit_it = FactoryBot.create :roster, name: 'Transit IT', created_at: Time.current.beginning_of_week(:friday)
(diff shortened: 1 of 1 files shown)


KernSharma:
ops now rotates twice daily at 4:30 am/pm (12-hour shifts over the same 5-week window IT covers), IT keeps the Friday–Friday weekly setup, and SKIP_ASSIGNMENTS is gone. Also tidied the admin block — memberships are created with admin set directly, which drops both SkipsModelValidations disables.

benmelz (approved):

PR #1112, opened 2026-07-09
Fix up assignment calendar display for ops

KernSharma:
Adds the three adjustments from #1109: nextDayThreshold '05:00' so shifts ending before 5 AM don't visually carry over, eventDisplay 'block' so same-day events render like multi-day ones instead of the dot format, and outline-style Open events (transparent background, bs-primary border/text) in place of the solid secondary fill.

Also sets displayEventTime false: with block display, same-day shifts would otherwise still show a "4:30a" time prefix while multi-day ones show nothing, which preserves the visual distinction the issue asks to remove. Easy to drop if the time text turns out to be wanted.

Closes #1109.


app/javascript/controllers/assignment_calendar_controller.js
```diff
@@ -22,6 +22,9 @@ export default class extends Controller {
       startParam: 'start_date',
       endParam: 'end_date',
       dayCellClassNames: 'day-empty',
+      nextDayThreshold: '05:00',
+      eventDisplay: 'block',
+      displayEventTime: false,
```
benmelz:
```suggestion
```
I think we definitely do want to display times still - it's implied that the previous day's assignment covers up to the displayed.

benmelz (changes requested):

benmelz (changes requested):
One more thing I thought about overnight - I also think that it isn't ideal that the "solid secondary" coloring of taken assignments is a bit bolder than the "border primary" coloring of open assignments.

Perhaps we change taken assignments to "border secondary" as well?

benmelz (approved):

PR #1115, opened 2026-07-12
Restrict assignment management to roster admins

KernSharma:
Restricts assignment CRUD to roster admins per #1113 — new/edit now fall through to manage, which drops the own-assignment exception. Members take shifts via the taker instead. Note destroy is included, so members can no longer give up their own shifts directly.

Flags from the wider audit: UserPolicy#create? does not pin entra_uid to the session (identity squatting); Twilio endpoints serve the on-call phone number unauthenticated (#772, intentional?); the taker lets members take already-assigned shifts.

Part of #1113.

### DIFF
--- a/app/policies/assignment_policy.rb
+++ b/app/policies/assignment_policy.rb
@@ -3,14 +3,7 @@
 class AssignmentPolicy < ApplicationPolicy
   authorize :roster, optional: true
 
-  def manage?
-    allowed_to?(:manage?,
-                record.roster) || (allowed_to?(:show?, record.roster) && record.user_id == user.id)
-  end
+  def manage? = allowed_to?(:manage?, record.roster)
 
   def index? = allowed_to?(:show?, roster)
-
-  def new? = allowed_to?(:show?, record.roster)
-
-  def edit? = allowed_to?(:show?, record.roster)
 end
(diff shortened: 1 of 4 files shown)


benmelz:
Could you elaborate on

> UserPolicy#create? does not pin entra_uid to the session (identity squatting);

?

KernSharma:
The create flow builds the new user from session values, but then user_params gets assigned over that, and user_params permits entra_uid. UserPolicy#create? checks that a session uid exists, that the requester doesn't have an account yet, and that admin isn't being set - but it never checks that the record's entra_uid actually matches the session. So someone who's authenticated through SSO but hasn't registered yet could submit a different uid and claim someone else's account before that person first signs in. The account would carry the submitter's phone and email, so calls for that person's shifts would forward to the wrong phone.

Simplest fix is either requiring record.entra_uid == session uid in create?, or not permitting entra_uid on create at all (I believe it's only there so admins can edit it later). Happy to do either here or as a follow-up.

benmelz:
Ah yes I see now.

> requiring record.entra_uid == session uid in create?

This one is more in line with how we currently do things. Let's include that in this PR.

benmelz (approved):

PR #1116, opened 2026-07-13
Stitch the AM/PM rosters to a single ops roster

zafnunhasnat-ops:
Part 2 #1103 
Adds a migration that merges the two temporary Transit Ops AM and Transit Ops EVE rosters back into a single Transit Operations roster.
It moves all memberships and assignments to the new roster and removes the two old rosters.
If a user belonged to both rosters, they keep one membership and remain an admin if they were an admin on either roster.
If two assignments end at the same time, one is moved forward by one second to avoid any database conflict.
I tested this by manually creating temporary AM/EVE rosters, memberships, and assignments in the local console, then ran the db:migrate / db:rollback commands against that data (not included in the commit).


db/migrate/20260710000000_stitch_am_pm_rosters.rb
```diff
@@ -0,0 +1,63 @@
...
+        end
+      end
+    end
+  end
+
+  def move_assignments(from:, to:)
+    from.each do |roster|
+      roster.assignments.to_a.each do |assignment|
+        end_datetime = assignment.end_datetime
+        end_datetime += 1.second while Assignment.exists?(roster: to, end_datetime: end_datetime)
```
benmelz: I think instead of trying to apply a sort of bandaid like this it's appropriate to just error out if there are duplicates.

benmelz (changes requested):
We can just edit the existing migration that's already responsible for some of the data paths (continuous assignments or something)

That one includes model stubs such that we will interface more directly with the tables instead of using whatever's in the application code (may reveal some problems with how we're accessing data),

PR #1123, opened 2026-07-14
ActiveAdmin demo: users table

KernSharma:
Demo of ActiveAdmin for admin controls, users table only, not intended to merge as-is. Uses the 4.0 beta since stable AA3 needs sprockets and we're on propshaft. Auth hooks into the existing Entra session and requires the app admin flag.

Assets now come precompiled from the activeadmin_assets gem (per review), so no tailwind, no extra build steps, and the node bump is reverted. The admin and entra_uid fields are intentionally read-only here since this resource bypasses action_policy; a real adoption would wire ActiveAdmin's authorization adapter to it. Also note the gem pulls in importmap-rails as a dependency (unused by the main app), and AA's controllers are detached from ApplicationController so verify_authorized doesn't reject them.


benmelz (changes requested):
One big thing that I think could make or break this whole project - I _really_ do not like the addition of tailwind and how we are being forced to build the provided active admin assets ourselves (i.e. the new rollup bundle and the new tailwind bundle).

Already it seems someone has already put up

https://github.com/jaynetics/activeadmin_assets/

that offers precompiled assets that we can just serve directly. I think it would be worth looking into that instead of explicitly bringing in a whole new css framework.

PR #1126, opened 2026-07-15
Add am/pm roster stitching to continuous assignment migration

zafnunhasnat-ops:
Refer to #1116 
Draft
Adds the roster stitching changes to the SetupContinuousAssignments migration.


benmelz:
Right this is already open as #1116

PR #1152, opened 2026-07-26
Upgrade FullCalendar to 7.0.2

werebus:
There are some additional compatibility changes that need to be made to #950 as well. I can take care of that PR, too.


app/assets/stylesheets/_fullcalendar.scss
```diff
@@ -0,0 +1,15 @@
...
+#assignment-calendar {
+  .calendar-title {
+    @include bootstrap.rfs(bootstrap.$h2-font-size);
+
+    font-weight: bootstrap.$headings-font-weight;
+  }
+
+  .calendar-day-empty {
+    cursor: pointer;
+  }
```
werebus, line 14: This is, arguably, an addition. When I was twiddling the JS, I remembered, "oh, right, you can click on empty days to add an assignment." But, that wasn't that obvious initially.

Now, the combination of `cursor: pointer`, and `window.location` manipulation is pretty rough, a11y-wise. But, given that #950 removes that feature, I'm not going to do _too_ much work on that front.

benmelz (approved):

PR #1161, opened 2026-07-28
Add support for multiple assignment generator definitions

zafnunhasnat-ops:
Closes #1129
Added support for multiple assignment definitions in AssignmentGenerator.
Updated the form to let users add and remove multiple rotations.
Added a Stimulus controller to handle adding and removing rotation fields.


db/migrate/20260612172421_setup_continuous_assignments.rb
```diff
@@ -73,6 +73,21 @@ def up
           curr.update!(end_datetime: curr.end_date + 1.day + roster.switchover.minutes)
           prev = curr
         end
+
```
benmelz: It looks like this all snuck back in (should be reverted back to how it is on `benmelz/assignments` currently)

benmelz (changes requested):

PR #1213, opened 2026-08-28
Use new shared workflows repo

benmelz:


public/400.html
```diff
@@ -1,18 +1,17 @@
 <!doctype html>
 
 <html lang="en">
-
```
werebus: I'm pretty confident that these are the Rails-generated error pages. Until we customize them (and therefor take on the maintenance of them) I think it makes sense to exclude them like we exclude config files from Rubocop.

werebus (changes requested):

werebus (approved):

